PR #2, opened 2016-03-11
Remove unused import

dlaboss:


spi/topology/src/main/java/quarks/topology/spi/graph/TWindowImpl.java
```diff
@@ -5,7 +5,6 @@
 package quarks.topology.spi.graph;
 
 import static quarks.window.Policies.alwaysInsert;
-import static quarks.window.Policies.processOnInsert;
```
ddebrunner: Just testing if a comment can be made!

PR #6, opened 2016-03-13
[QUARKS-18] [WIP] Split function based upon an enumeration

Cazen:
I've created PR for newbie's issue but afraid to there are mistake or misunderstood.

Please feel free to advise me.


ddebrunner:
Thanks for taking this on:

The idea was for the splitter function to have the signature `Function<T,E>` where E is the enum type, so that the application developer just uses enum constants rather than integers.

For example if one was splitting on the severity of a log event and the log event class has a `Severity getSeverity()` method where Severity is an enum, then the split would look something like:

`List<TStream<LogEvent>>  splits = logEvents.split(Severity.class, e -> e.getSeverity());`

https://issues.apache.org/jira/browse/QUARKS-18

ddebrunner:
I also think it's possible to implement this by having the new split method that takes an enum just call the  existing split method, using a function that wraps the user's enum split function and returns the enum's ordinal().

ddebrunner:
I wonder if this enum split should return:

`List<TStream<T>>` // like the existing split, the list would be ordered by the enum ordinals

`EnumMap<E,TStream<T>>`  // possibly more natural for enums since ordinal() does say ' Most programmers will have no use for this method.'

Example use of first (with an example Severity enum of HIGH, LOW)

```
TStream<LogEvent> highStream = splits.get(HIGH.ordinal());
TStream<LogEvent> lowStream = splits.get(LOW.ordinal());
```

I think the map approach cleans this up to:

```
TStream<LogEvent> highStream = splits.get(HIGH);
TStream<LogEvent> lowStream = splits.get(LOW);
```

Cazen:
Good day Debrunner thank you for your pointing
I will check it and modify PR after company work
Have a good day

PR #7, opened 2016-03-14
[QUARKS-10] Iot provider

ddebrunner:
A provider that registers services to allow:

* PublishSubscribe Service
* Application Service
* JSON control service
* use of a single IoTDevice by multiple applications through publish-subsribe
* Integration of device commands with identifier `quarksControl`

### DIFF
--- a/apps/iot/src/test/java/quarks/test/apps/iot/EchoIotDevice.java
+++ b/apps/iot/src/test/java/quarks/test/apps/iot/EchoIotDevice.java
@@ -21,6 +21,7 @@ package quarks.test.apps.iot;
 
 import static quarks.function.Functions.discard;
 
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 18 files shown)


apps/iot/src/test/java/quarks/test/apps/iot/EchoIotDevice.java
```diff
@@ -78,7 +78,7 @@ private static String getCommandIdFromEvent(String eventId, JsonObject evPayload
         if (evPayload.has(EVENT_CMD_ID))
             return evPayload.getAsJsonPrimitive(EVENT_CMD_ID).getAsString();
         else
-            return "ec_" + eventId;
+            return eventId;
     }
```
dlaboss: this applies to `commands()` on line 110...
- in the case where there's no cmd filtering, why is it desirable/necessary to pass `echoCmds` through a `alwaysTrue` filter?  Why not just return `echoCmds`?
- can't multi-cmd case processing be simplified to only have a single `filter` and no `union`... and merged with the single cmd filtering case for that matter? e.g.,
```
         Set<String> cmds = new HashSet<>(Arrays.asList(commands));
         return echoCmds.filter(j -> cmds.contains(j.getAsJsonPrimitive(CMD_ID).getAsString());
```

ddebrunner: Good points.

PR #11, opened 2016-03-15
[QUARKS-18] Split function based upon an enumeration

Cazen:
I'm very sorry but there are conflict error on my previous PR.
I tried to solve it but looks hard :( .. So recreate PR
Please refer : #6 Previous Discussion

### DIFF
--- a/api/topology/src/main/java/quarks/topology/TStream.java
+++ b/api/topology/src/main/java/quarks/topology/TStream.java
@@ -4,17 +4,14 @@
 */
 package quarks.topology;
 
+import quarks.function.*;
+import quarks.oplet.core.Pipe;
+import quarks.oplet.core.Sink;
+
+import java.util.EnumMap;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-
-import quarks.function.Consumer;
-import quarks.function.Function;
-import quarks.function.Predicate;
-import quarks.function.ToIntFunction;
-import quarks.function.UnaryOperator;
-import quarks.oplet.core.Pipe;
-import quarks.oplet.core.Sink;
 
 /**
  * A {@code TStream} is a declaration of a continuous sequence of tuples. A
@@ -195,6 +192,18 @@ public interface TStream<T> extends TopologyElement {
      *             if {@code n <= 0}
      */
     List<TStream<T>> split(int n, ToIntFunction<T> splitter);
+
+    /**
+     * Split a stream's tuples among {@code enumClass.size} streams as specified by
+     * {@code splitter}.
+     *
+     * @param enumClass
+     *            enum data to split
+     * @param splitter
+     *            the splitter function
+     * @return EnumMap<E,TStream<T>>
+     */
+    <E extends Enum<E>> EnumMap<E,TStream<T>> split(Class<E> enumClass, Function<T, E> splitter);
 
     /**
      * Declare a stream that contains the same contents as this stream while
--- a/api/topology/src/main/java/quarks/topology/TStream.java
+++ b/api/topology/src/main/java/quarks/topology/TStream.java
@@ -4,7 +4,11 @@
 */
 package quarks.topology;
 
-import quarks.function.*;
+import quarks.function.Consumer;
+import quarks.function.Function;
+import quarks.function.Predicate;
+import quarks.function.ToIntFunction;
+import quarks.function.UnaryOperator;
 import quarks.oplet.core.Pipe;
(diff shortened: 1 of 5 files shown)


spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
```diff
@@ -96,6 +89,35 @@ protected Graph graph() {
...
+    public <E extends Enum<E>> EnumMap<E,TStream<T>> split(Class<E> enumClass, Function<T, E> splitter) {
+
+        E[] es = enumClass.getEnumConstants();
+
+/*
+        List<TStream<T>> outputs = split(es.length, e -> IntStream.range(0, es.length).filter(i-> es[i].equals(splitter.apply(e)))
+            .mapToObj(i -> es[i].ordinal()).findAny().orElse(-1));
+*/
+
+        List<TStream<T>> outputs = split(es.length, new ToIntFunction<T>() {
```
ddebrunner: The function is incorrect as it can call the splitter function multiple times, I think it can be simplified to something like:

```
t -> {E split = spitter.apply(t); return split != null ? split.ordinal() : -1;}
```

which also implies that when the splitter returns null that the tuple is discard which should be documented in the api method's javadoc.

spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
```diff
@@ -96,6 +89,35 @@ protected Graph graph() {
...
+            @Override
+            public int applyAsInt(T input) {
+                for(int i = 0; i < es.length; i++){
```
dlaboss: I was also going to comment on simplifying the fn.  And (then eliminating the need for the intermediate "es" variable) I think the returnMap construction could be simplified to:
```
for (E e : enumClass.getEnumConstants()) {
    returnMap.put(e, outputs.get(e.ordinal());
}
```

api/topology/src/main/java/quarks/topology/TStream.java
```diff
@@ -4,18 +4,15 @@
 */
 package quarks.topology;
 
+import quarks.function.*;
```
dlaboss: We don't have a formal coding style doc but the rest of the code uses explicit imports instead of wildcards.  Generally, it's best to avoid arbitrary changes when introducing new functionality.  Generally, nice work btw.

Cazen:
I changed my PR on the advice. Thank you for help!

- Using explicit import instead of starred
- Changing error code in ConnectorStream
- Adding testcase for 0-size enum

@dlaboss I really appreciate your help! And could I ask one more question?
I tried to remove "es" variable, but it will make enumClass.getEnumConstants() call twice in the code.
(Additionaly, I've added "es==null" check) Is there any good way to solve this?

spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
```diff
@@ -96,6 +97,28 @@ protected Graph graph() {
     }
 
     @Override
+    public <E extends Enum<E>> EnumMap<E,TStream<T>> split(Class<E> enumClass, Function<T, E> splitter) {
+
+        E[] es = enumClass.getEnumConstants();
+        if(es == null) {
```
ddebrunner: I'm merging this pull request but I'm not sure I understand this code:

E is declared as an enum so the message the exception throws is not valid.

I see there is a test `testSplitWithEnumForZeroSizeClass` , is that what this check is for? (Nice job adding the test, I didn't realize that there could be empty enums).

If the check is for that then maybe the message should be changed? Just from reading the code I'd be tempted to remove the check as it appears as though it can't happen since E is always an enum.

Cazen:
@ddebrunner Sorry for confusing. I have reviewed the code once again, realize that you are right.

1. I added testcase(`testSplitWithEnumForZeroSizeClass`) for detecting 0 size enum. It sounds weird but would be helpful i thought.
2. Checking `es == null` is unnecessary as you mentioned. Yes It declared as an enum. I missed. Is there any chance to correct?

Thank you for closely review for my code. I know all committers are busy and contribute to Quarks using personal time. I do not attempt to interfere but there are a lot of mistake that waste of your time. I promise I will work more carefully next time.

Cheers!

ddebrunner:
@Cazen Thanks for contributing, don't worry, I didn't waste my time, I also got to learn something new, there can be enums with no values!

You can always "correct" any code by a new pull request, I tend to think any code can always be improved, no-one writes perfect code on the first attempt!

PR #14, opened 2016-03-15
[QUARKS-16] [WIP] Add topology sample for combining streams processing results

queeniema:
Purpose: Applying different processing against a set of streams and combining the resulting streams into a single stream.

- Polls a simulated heart monitor to periodically obtain blood pressure readings.
- Splits the readings by blood pressure category into separate streams.
- Applies different processing on each stream to generate alert streams.
- Combines the alert streams into a single stream and prints the alerts.

### DIFF
--- a/samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
+++ b/samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
@@ -1,8 +1,21 @@
 /*
-# Licensed Materials - Property of IBM
-# Copyright IBM Corp. 2015,2016
-*/
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
 
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
 package quarks.samples.topology;
 
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


ddebrunner:
@queeniema Have you submitted an ICLA to the ASF?

queeniema:
@ddebrunner Yes, I have.

samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
```diff
@@ -0,0 +1,158 @@
+/*
```
ddebrunner: Header for new files needs to be ASF licence header:

http://apache.org/legal/src-headers.html#headers

queeniema: Should the IBM copyright be removed in favor of the ASF license header? Or should the license be added in addition to the copyright?

samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
```diff
@@ -0,0 +1,158 @@
...
+                .map(tuple -> { return "Emergency! See to patient immediately!\n"; })
+                .tag("hypertensive")
+                .modify(tuple -> tuple.toUpperCase())
+                .tag("hypertensive")
+                .modify(tuple -> "Hypertensive Crisis!!!\n" + tuple)
+                .tag("hypertensive");
+
+        // Union two streams to obtain a single stream containing alerts from the normal and
+        // prehypertension alert streams
+        TStream<String> normalAndPrehypertensionAlerts = normalAlerts.union(prehypertensionAlerts);
```
ddebrunner: Just curious if there is some point you are trying to get across with the multiple unions, rather than the equivalent single union?

queeniema: To show that `union()` is overloaded. I wanted to demonstrate both cases as this was originally intended as a recipe.

samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
ddebrunner: These are new files you are contributing to the ASF under the ICLA, thus they only have the ASF licence header.

samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
ddebrunner: Maybe some comment about that, otherwise a reader might be wondering why it was done this way. Maybe something along the lines of additional processing for these streams could go here.

ddebrunner:
Nice sample, like the use of tags.

queeniema:
Thanks! I committed the changes. If there are any other suggestions/feedback, please feel free to comment.

PR #13, opened 2016-03-15
Add info about using github

dlaboss:

PR #15, opened 2016-03-16
Adjust "github use" doc based on feedback.

dlaboss:
See PR #13 for feedback.
This is in a new PR because PR #13 , while it doesn't yet appear so, was already merged.


ddebrunner:
FYI - the bullet list for the summary of the github fork/clone/commit/pull reaquest seems to have messed up markdown, it's appearing as a single paragraph.

PR #24, opened 2016-03-17
[QUARKS-40] remove non-existent src/test/java dir reference from .classpath

dlaboss:


dlaboss:
travis failure is unrelated:  
WebSocketClientTest.testReconnect() failed.  May be intermittent.
I'll merge and then investigate.

PR #25, opened 2016-03-17
[QUARKS-16] Update split to use lambda function

queeniema:
Change made to avoid using `serialVersionUID`.

### DIFF
--- a/samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
+++ b/samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
@@ -25,7 +25,6 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import quarks.console.server.HttpServer;
-import quarks.function.ToIntFunction;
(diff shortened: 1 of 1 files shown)


samples/topology/src/main/java/quarks/samples/topology/CombiningStreamsProcessingResults.java
```diff
@@ -62,32 +61,28 @@ public static void main(String[] args) throws Exception {
...
-                    return 3;
-                } else if (tuple.get("Systolic") >= 180 && tuple.get("Diastolic") >= 110)  {
-                    // Hypertensive Crisis
-                    return 4;
-                } else {
-                    // Invalid
-                    return -1;
-                }
+        List<TStream<Map<String, Integer>>> categories = readings.split(6, tuple -> {
+            if (tuple.get("Systolic") < 120 && tuple.get("Diastolic") < 80) {
```
ddebrunner: Unrelated to this change but it jumps out at me that the code might be clearer and more compact if you fetched the two values into local variables first and then used those variables rather than getting each value multiple times.

PR #29, opened 2016-03-19
[QUARKS-47] Remove unnecessary null check in ConnectorStream.split(enumClass, splitter)

Cazen:


Cazen:
@ddebrunner Good day! Could you please review this change?
It related PR #11.

wmarshall484:
+1 to merge, E is already defined as an Enum, so the null check is unnecessary

wmarshall484:
Hm, odd that asf-git didn't close this automatically.

Cazen:
Thank you @wmashall484
Have a good day!

PR #33, opened 2016-03-21
[QUARKS-52] fix test keystore path formation on windows

dlaboss:


connectors/wsclient-javax.websocket/src/test/java/quarks/tests/connectors/wsclient/javax/websocket/KeystorePath.java
```diff
@@ -5,22 +5,25 @@
...
-        path += "/wsclient-javax.websocket/src/test/keystores/" + storeLeaf;
-        if (!new File(path).exists())
+        // Get the path to the <repo>/connectors
+        Path path = new File(pathStr).toPath();
+        do {
+            if (path.endsWith("connectors"))
+                break;
+            path = path.getParent();
+        } while (path.getNameCount() > 0);
+        path = path.resolve("wsclient-javax.websocket/src/test/keystores/" + storeLeaf);
```
ddebrunner: This will still be broken under Windows I think because the path separator is hard-coded for linux.

dlaboss: All is well as @kmarsden verified in https://issues.apache.org/jira/browse/QUARKS-52.  Windows is typically generous accepting both forms of separators.

ddebrunner: Still, portable Java code should not assume '/' will work as a file separator. File.separatorChar.

PR #44, opened 2016-03-23
[QUARKS-67] skip public server tests if can't connect

dlaboss:


connectors/wsclient-javax.websocket/src/test/java/quarks/tests/connectors/wsclient/javax/websocket/WebSocketClientTest.java
```diff
@@ -678,6 +681,24 @@ public void testSslClientAuthNeg() throws Exception {
...
+            int port = uri.getPort();
+            if (port == -1)
+                port = uri.getScheme().equals("ws") ? 80 : 443;
+            Socket s = new Socket();
+            s.connect(new InetSocketAddress(uri.getHost(), port), 5*1000/*cn-timeout-msec*/);
+            s.close();
+        } catch (Exception e) {
+            System.err.println("Unable to connect to WebSocket server "+wsUri+" : "+e.getMessage());
+            e.printStackTrace();
+            assumeTrue(false);
```
ddebrunner: Is it possible to use the overload that takes a message that will be reported in the junit report, or does that require a newer version of JUnit?

dlaboss: Yup, those overloads seem to have been added in added in 4.11.  
DEVELOPMENT.md says only 4.10 is required.

PR #48, opened 2016-03-25
[QUARKS-78] Range analytic utility

dlaboss:
- promote samples.apps.Range to quarks.analytics.sensors
- see Range javadoc for Guava Range issues and such

### DIFF
--- a/analytics/sensors/src/main/java/quarks/analytics/sensors/Range.java
+++ b/analytics/sensors/src/main/java/quarks/analytics/sensors/Range.java
@@ -135,16 +135,15 @@ public final class Range<T extends Comparable<?>> implements Predicate<T>, Seria
     public static enum BoundType {/** exclusive */ OPEN, /** inclusive */ CLOSED};
     
     /**
-     * Create a new Range<T>
+     * Create a new Range<T>.  Private like Guava Range.
      * <p>
      * See {@link Ranges} for a collection of convenience constructors.
-     * 
-     * @param lbt {@link BoundType} for the lowerEndpoint
      * @param lowerEndpoint null for an infinite value (and lbt must be OPEN)
+     * @param lbt {@link BoundType} for the lowerEndpoint
      * @param upperEndpoint null for an infinite value (and ubt must be OPEN)
      * @param ubt {@link BoundType} for the upperEndpoint
      */
-    public Range(BoundType lbt, T lowerEndpoint, T upperEndpoint, BoundType ubt) {
+    private Range(T lowerEndpoint, BoundType lbt, T upperEndpoint, BoundType ubt) {
         this.lowerEndpoint = lowerEndpoint;
         this.upperEndpoint = upperEndpoint;
         this.lbt = lbt;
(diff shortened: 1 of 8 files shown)


ddebrunner:
+1 for adding this, not sure why I would care about Guava Range in the Javadoc, that's more an implementation note.

Javadoc for methods could be improved, e.g:
```
/** (a..*) (exclusive) */
    public static <T extends Comparable<?>> Range<T> greaterThan(T v) {
```

References 'a' but 'a' is not a parameter.

ddebrunner:
Does it make sense for Range to implement Predicate?

A couple of simple examples of its use would be helpful, got somewhat lost in the full application from a quick scan.,

dlaboss:
Yup, I'll improve the javadoc.

Re: should it implement Predicate?  I think the answer is "no".  I believe Range isa Predicate is only useful if the stream's tuples are "primitive" types (e.g., Integer).  In the probably very rare case where there is such a stream, lambda expressions make it trivial for the user to just supply `tuple -> myRange.contains(tuple)` (instead of just `myRange`).   Fwiw, while Guava Range implements its Predicate, the test() override is documented as deprecated and says callers should just use "contains()".  Seems like they now regret it being a Predicate.  Actually, for a short time I had this Range implement quarks Predicate and removed it :-)

ddebrunner:
@dlaboss Could you expand on why it's not useful when the tuples are not wrapper classes?

dlaboss:
If "simple" streams like `TStream<Integer>` aren't likely then this seemed unnecessary:
```
TStream<Integer> s = ...
TStream<Integer> filtered = s.filter(myRange);  // vs t -> myRange.contains(t)
```
Hmm... I guess I hadn't considered a case like Filters.deadband() where a "simple type" Predicate must be supplied for a type already extracted from a more complex tuple.  e.g.,
```
TStream<JsonObject> readings = ...
Range<Double> deadbandRange = Range.closed(10.0d, 30.0d);
TStream<JsonObject> filtered = Filters.deadband(readings,
                           tuple -> tuple.get("reading").getAsDouble(),
                           deadbandRange);    // vs   v -> deadbandRange.contains(v)
```
@ddebrunner is that sufficient rationale for making Range a Predicate?  Including in light of Guava Range's seemingly regret about that [here](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Range.java#L436)?

ddebrunner:
@dlaboss Maybe I'm misunderstanding what Range can be used for. Can Range be used on a arbitrary Java class that implements Comparable? If so then why is Range as a Predicate not useful in that case?

dlaboss:
OK you caught me :-)  As was noted in the javadoc the initial implementation only supported a subset of Comparables: Integer,Long,Short,Byte,Float,Double,BigInteger,BigDecimal.  I've revisited things to remove that restriction.  

I've also added "implements Predicate".  Can you provide a sample  / pseudo code for a non-trivial case where and how you want to be using that fact so I can ensure that need is met.

I'll be pushing changes for the above and other issues raised shortly.

dlaboss:
I'll merge this tomorrow afternoon if there are no other comments... and address things after if things come in after that.

PR #53, opened 2016-03-30
[ASK] Is it possible to change an Author?

Cazen:
Good day. I've found that my recent commit's email address is "magic12#"(Unknown)
It's my mistake. In fact, this is my password and I've changed it immediately.
So, if allowed, I would like to change the author email address of previous commit to my original address. But I'm not sure can be alowed of such type of PR.
Could I ask to review it? Of course, If it does not permitted, I will close this PR
Sorry for confusing

### DIFF
--- a/analytics/math3/build.xml
+++ b/analytics/math3/build.xml
@@ -1,4 +1,20 @@
-<project name="quarks.analytics.math3" default="all" 
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements. See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership. The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License. You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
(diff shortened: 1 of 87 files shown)


ddebrunner:
Given the warning here:
https://help.github.com/articles/changing-author-info/

I'm not sure we should change the commit author for the two commits that have this issue.

I'm also not sure what this pull request does since it shows diffs that have already been applied, so unclear to me the affect it would have on the repo history. What git commands were used to generate it?

ddebrunner:
@Cazen It's great to have you active in the community, as you hopefully become more active you should consider filing an Individual Contributor License Agreement  (ICLA) with Apache Software Foundation, if you haven't already. I looked a while ago and didn't see one listed for you. The ICLA would be required for a committer.

See: http://quarks.incubator.apache.org/docs/community

FYI  - your existing commits have the e-mail listed as `Cazen@korea.com` rather than the e-mail displayed for the @Cazen account, which is also different to the e-mail used to post to the list.

Cazen:
Good Day @ddebrunner I've already seen that pages you mentioned and generated this PR under that command. 
I also read WARNING phrase. So I think it's better that do not allow the change author operation unless anyone have previous experience like this situation. Since it is the first time to me involved in a large-scale opensource project using git, I was wondering if this type of operation is permitted.
And thank you for your consideration. My email address both cazen.lee@gmail.com and cazen@korea.com are linked with github. I will fill the ICLA form and submit soon.

PR #56, opened 2016-03-30
[QUARKS-64] Merge Hello Quarks! recipe into HelloWorld sample

queeniema:


ddebrunner:
Need to update the scripts as well.

PR #59, opened 2016-03-31
[WIP] QUARKS-66 Job monitoring application

vdogaru:
Monitor application listens on JobRegistry events and resubmits unhealthy jobs. The monitored applications must be registered with an ApplicationService prior to submission, otherwise the monitor application cannot restart them. Packaged in new module quarks.apps.runtime.jar.

The monitor application depends on the following service changes:
- ApplicationService provides the names of registered applications
- ControlService.queryInterfaces returns the registered controls which implement a specified interface.  Used by the monitor application to get ApplicationServiceMXBean control in order to restart failed applications.


apps/runtime/build.xml
```diff
@@ -0,0 +1,29 @@
+<project name="quarks.apps.runtime" default="all"
```
ddebrunner: Missing ASF licence header.

api/execution/src/main/java/quarks/execution/services/ControlService.java
```diff
@@ -91,4 +93,18 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+     * {@linkplain ControlService#registerControl(String, String, String, Class, Object) register}
+     * the control.
+     * 
+     * @param controlInterface
+     *              Public interface identifying the controls to be retrieved. 
+     * @return a set containing the controls registered with the given 
+     *              interface. If no control satisfies the query, an empty 
+     *              set is returned.
+     */
+    <T> Set<T> queryInterfaces(Class<T> controlInterface);
```
ddebrunner: It's not a 'query', it's just a 'get' (read its description). Rename to 'getControls' or 'getControlMBeans'??

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * Submits an application using an {@code ApplicationServiceMXBean} control 
+     * registered with the specified {@code ControlService}.
+     * 
+     * @param applicationName the name of the application to submit
+     * @param controlService the control service
+     */
+    public static void submitApplication(String applicationName, ControlService controlService) {
```
ddebrunner: Not sure I understand the purpose of this method. If code can get the ControlService it can get the application service.

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+    
+    /**
+     * Submits the application topology.
+     * 
+     * @return the job.
+     * @throws InterruptedException
+     * @throws ExecutionException
+     */
+    public Job submit() throws InterruptedException, ExecutionException {
```
ddebrunner: I think this indicates there needs to be some standardization around applications, so they don't each have to implement their own submit method. Thinking that maybe there is an interface applications can implement to provide consistency, and we encourage jobs to start through the ApplicationServiceMXBean, and not by themselves.

I'll put together some proposal, it's wouldn't block this being merged.

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
```diff
@@ -0,0 +1,204 @@
...
+     * @param name the topology name
+     * @return the application topology
+     */
+    protected Topology declareTopology(String name) {
+        Topology t = provider.newTopology(name);
+        TStream<JsonObject> jobEvents = JobEvents.source(
+                t, 
+                (evType, job) -> { return MonitorAppEvent.toJsonObject(evType, job); }
+                );
+        jobEvents = PlumbingStreams.isolate(jobEvents, true);
```
dlaboss: Isn't the isolate superfluous?  JobEvents.source() just wraps Topology.source() and that defines the stream is isolated.  Probably good to add the same doc to JobEvents.source().

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        @Override
+        public void accept(JsonObject value) {
+            ControlService controlService = rts.get().getService(ControlService.class);
+            JsonObject job = MonitorAppEvent.getJob(value);
+            String applicationName = MonitorAppEvent.getJobName(job);
+
+            logger.info("Will restart monitored application {}, cause: {}", applicationName, value);
+            submitApplication(MonitorAppEvent.getJobName(job), controlService);
```
dlaboss: replace 2nd call to `getJobName()` with `applicationName`?  Not an efficiency concern, just makes things clearer.

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
```diff
@@ -0,0 +1,204 @@
...
+            Set<ApplicationServiceMXBean> controls = 
+                    controlService.getControls(ApplicationServiceMXBean.class);
+            if (controls.isEmpty()) {
+                throw new IllegalStateException(
+                        "Could not find a registered control with the following interface: " + 
+                        ApplicationServiceMXBean.class.getName());                
+            }
+            for (ApplicationServiceMXBean control : controls)
+// TODO add ability to submit with the initial application configuration
+                control.submit(applicationName, null);
```
dlaboss: Just trying to understand / imagine when there might be more than one ApplicationServiceMXBean registered and then what it might mean to submit the same applicationName multiple times once for each AS control?

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
ddebrunner: > Topology.source() and that defines the stream is isolated
Not sure what that means? How does it define it as isolated?

dlaboss: Sorry for the confusion.  I meant wraps `Topology.events()` and its doc defines the stream as isolated [here](https://github.com/apache/incubator-quarks/blob/master/api/topology/src/main/java/quarks/topology/Topology.java#L116)  And the impl GraphTopology.events() already returns a PlumbingStreams.isolate()'d stream :-)

api/topology/src/main/java/quarks/topology/services/ApplicationService.java
```diff
@@ -63,4 +65,11 @@ Licensed to the Apache Software Foundation (ASF) under one
      * @see ApplicationServiceMXBean
      */
     void registerTopology(String applicationName, BiConsumer<Topology, JsonObject> builder);
+
```
dlaboss: Above for `registerTopology()` what are the requirements for the appName?  What happens if one by that name is already registered?  By definition elsewhere, is a Job's appName already required to be unique? Regardless, seems like it could help to add doc here to clarify things.

I'm also wondering about this "register with appName prior to submit" model vs say "register the *Job* following the submit".  A post submit registration scheme seems to enable leaving it to the system/provider-impl to decide what to use as an identifier to find the topology-builder to rebuild/resubmit the job.   It also feels more logical to express "I want this Job monitored" rather than "I want a/all jobs with this appName monitored"... though maybe that's just me.  Does the pre-submit scheme handle recovery from certain startup failures that the post-submit scheme can't?

Is there also a need for an unregisterTopology() or is it just that an explicitly cancelled job is effectively automatically unregistered?

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
ddebrunner: Ahh, thanks for the clarification.

api/topology/src/main/java/quarks/topology/services/ApplicationService.java
ddebrunner: I'm solving a concrete use case, where the device has a number of pre-installed applications but not all are running. Remotely an application can be enabled, for example when a customer pays for a specific feature/service, the device provider turns on that application and presumably events from the device start flowing back to a central system to provide the service to the customer.

If there are other use cases that need to be supported additional functionality can be added, maybe the discussion could move out of this pull request.

apps/runtime/src/main/java/quarks/apps/runtime/MonitorApp.java
vdogaru: This and your next comment resolved in https://github.com/apache/incubator-quarks/pull/65

api/topology/src/main/java/quarks/topology/services/ApplicationService.java
dlaboss: Ah, I see how a register-after-submit wouldn't work for that case.  I was basing my comment/question on the just the "monitoring" and "restart failed jobs" use case in the title, etc.

PR #60, opened 2016-03-31
Added join and joinLast functionality.

wmarshall484:


api/topology/src/main/java/quarks/topology/TStream.java
```diff
@@ -439,4 +440,11 @@ else if (WARNING.equals(lr.getLevel()))
      * @return set of tags
      */
     Set<String> getTags(); 
+    
+    <J, U, K> TStream<J> join(Function<T, K> keyer, TWindow<U, K> window, BiFunction<T, List<U>, J> joiner);
+    
+    <J, U, K> TStream<J> joinLast(Function<T, K> keyer, TStream<U> lastStream, Function<U, K> lastStreamKeyer, BiFunction<T, U, J> joiner);
+
+    <J, U> TStream<J> joinLast(TStream<U> lastStream, BiFunction<T, U, J> joiner);
```
ddebrunner: Why have this method? We don't provide unpartitioned variants of other methods do we?

spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
```diff
@@ -163,6 +175,100 @@ protected Graph graph() {
     }
     
     @Override
+    public <J, U, K> TStream<J> join(Function<T, K> keyer,
+            TWindow<U, K> twindow, BiFunction<T, List<U>, J> joiner) {
+        
+        if(twindow instanceof TWindowImpl){
+            TStream<U> lastStream = twindow.feeder();
```
ddebrunner: There's several lines of code that are common between the two implementations, can they be shared?

spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
```diff
@@ -163,6 +175,100 @@ protected Graph graph() {
...
+            TimeUnit unit = ((TWindowTimeImpl<U, K>)(twindow)).getUnit();
+            Window<U, K, InsertionTimeList<U>> window =
+                    Windows.window(
+                            alwaysInsert(),
+                            scheduleEvictIfEmpty(time, unit),
+                            evictOlderWithProcess(time, unit),
+                            processOnInsert(),
+                            twindow.getKeyFunction(),
+                            insertionTimeList());
+            Aggregate<U,Object,K> op = new Aggregate<U,Object,K>(window, processor);
```
ddebrunner: Can some explanation be added, it's unclear why an Aggregate is used that is unconnected.

I guess i was expecting a new Join oplet to be created & used.

spi/topology/src/main/java/quarks/topology/spi/graph/TWindowTimeImpl.java
```diff
@@ -94,5 +94,33 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+    /**
+     * @return the time
+     */
+    public long getTime() {
+        return time;
+    }
+
+    /**
+     * @param time the time to set
+     */
```
dlaboss: windowing impl newbie wondering about the addition of the setters...  the ctor requires the values be supplied.  Is it really necessary to make them changeable post construction?  They can't be made final?  Changing the values after calling aggregate() or batch() seems to have no affect on the behavior of the windows associated with them.  Maybe just WIP, maybe just need more doc stating that, maybe some class doc clarifying usage of the class - e.g., is a TWindowTimeImpl instance supposed to represent a single window instance or a factory for one or more actual windows (i.e., it's legal to call both aggregate() and batch(), or each one more than once)?

api/window/src/main/java/quarks/window/Window.java
```diff
@@ -157,4 +158,10 @@ Licensed to the Apache Software Foundation (ASF) under one
      */
     Function<T, K> getKeyFunction();
 
+    /**
+     * Retrieves the partitions in the window.
+     * @return A map of the window's keys and partitions.
```
dlaboss: How about adding doc for the behavior / use requirements of the returned map as they relate to: is the returned map/partitions "stable" or can the state change? What happens if one attempts to change the map or the partitions it contains - is it backed by the live window's objects?

spi/topology/src/main/java/quarks/topology/spi/graph/TWindowTimeImpl.java
wmarshall484: The set() methods were generated accidentally; they have been removed.

api/window/src/main/java/quarks/window/Window.java
wmarshall484: The returned map is stable when synchronizing on its intrinsic lock, for example:
```java
Map<Integer, Partition<Integer, Integer, ArrayList<Integer> > > partitions = window.getPartitions();
synchronized(partitions){
    // Stable operations on/with partitions
}
```
I've amended the doc to address this.

spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
wmarshall484: Fixed.

spi/topology/src/main/java/quarks/topology/spi/graph/ConnectorStream.java
wmarshall484: The Aggregate isn't unconnected. With the next line
```java
lastStream.pipe(op);
```
 its state is set to update based on the tuples from `twindow.feeder` stream. The `window` object which was created earlier is included in the `joiner` function via closure so its partitions can be accessed.

Including the window in the `joiner` function through a closure removes the need to create an additional oplet with two input ports. The downside is that this approach only works if the Quarks application is only running inside of a single JVM, which it is for now (but maybe not later?)

api/window/src/main/java/quarks/window/Window.java
dlaboss: Maybe you've also made adjustments but on lines 188 & 216 below in ConnectorStream, it's not synchronizing the partitions map when getting the partition of interest.  Q: would using/returning a synchronized map be better?

PR #64, opened 2016-03-31
QUARKS-99 Improve Javadoc for IotfDevice

ddebrunner:


connectors/iotf/src/main/java/quarks/connectors/iotf/IotfDevice.java
```diff
@@ -69,11 +109,56 @@ public IotfDevice(Topology topology, Properties options) {
...
+     */
     public IotfDevice(Topology topology, File optionsFile) {
         this.topology = topology;
         this.connector = new IotfConnector(optionsFile);
     }
     
+    /**
+     * Create an {@code IotfDevice} connector to the Quickstart service.
+     * Quickstart service requires no-sign up to use, but only supports
+     * device commands for a single
```
dlaboss: "...for a single" ?

ddebrunner: Good catch!

PR #66, opened 2016-04-01
QUARKS-106 Replace getControls() with getControl()

ddebrunner:


runtime/jmxcontrol/src/main/java/quarks/runtime/jmxcontrol/JMXControlService.java
```diff
@@ -127,29 +126,40 @@ public void unregister(String controlId) {
...
+    public <T> T  getControl(String type, String alias, Class<T> controlInterface) {
         try {
             MBeanServer mBeanServer = getMbs();
-            Set<ObjectName> names = getObjectNamesForInterface(controlInterface.getName());
+            Set<ObjectName> names = getObjectNamesForInterface(type, alias, controlInterface.getName());
+            
+            if (names.isEmpty())
+                return null;
+            if (names.size() != 1)
+                throw new RuntimeException("Alias " + alias + " not unique for type " + type);
```
vdogaru: I'm thinking that I would expect JMXControlService.register to throw an exception if I register two controls with the same type and alias (according to the ControlService docs?) Instead, I get an exception later, when I query?

ddebrunner: I'm just changing getControls() in this pull request. It only expects a single MBean registered, so if it finds more it complains.

PR #67, opened 2016-04-01
Rework IotfSensors to separate Heartbeat functionality

ddebrunner:


samples/connectors/src/main/java/quarks/samples/connectors/iotf/IotfSensors.java
```diff
@@ -109,6 +99,31 @@ public static void simulatedSensors(IotDevice device, boolean print) {
...
     }
+    
+    /**
+     * Create a heart beat device event with
+     * identifier {@code heartbeat} to
+     * ensure there is some immediate output and
+     * the connection to IoTF happens as soon as possible.
+     * @param device IoT device
+     */
+    public static void heartBeat(IotDevice device, boolean print) {
```
dlaboss: Wondering about promoting (unncessary?) heartbeat messages.  While IotDevice isn't necessarily MQTT specific, MQTT has inherent mechanisms for dealing with this sort of thing:  KeepAlive and LastWillAndTestament.  Discussions [here-on-stackoverflow](http://stackoverflow.com/questions/17270863/mqtt-what-is-the-purpose-or-usage-of-last-will-testament) and [here-on-hivemq](http://www.hivemq.com/blog/mqtt-essentials-part-9-last-will-and-testament). Perhaps IotDevice should include an interface for expressing this and impls do what's needed to comply?

ddebrunner: Any thoughts on how this sample method comment could be improved to not promote its use? It does already state what the purpose of the heartbeat is for this sample.

Not also that a heartbeat device event performs a somewhat different role to KeepAlive/LastWillAndTestament in that subscribers have visibility to the "I'm still connected" events

dlaboss: I guess the comment could just add something like "If you are using an MQTT-based IotDevice, you may want to consider using the MQTT KeepAlive and LastWillAndTestament mechanisms instead of an explicit heartbeat."  ?

It's true that this heartbeat provides a visible last-heartbeat-timestamp to clients.  However, a visible&reliable "I'm still connected" status seems to be possible using just the KA/LWT.  

That HiveMQ reference suggests: with a "retained" LWT to a topic "status" with the value "offline", the device initially sends a *single* retained msg "online" to that topic.  Clients will see/receive "online" until the device is no longer connected, at which point they will receive "offline" (by virtue of the KA-induced LWT).  I think there's also an implied: upon graceful shutdown, the device sends a single retained "offline" to "status".

Regardless, should an issue be created to discuss (or track) adding something to IotDevice for this higher level concept/feature, and to the MqttDevice and IotfDevice impls? Don't know if something like this is already part of the Watson IoT framework.

ddebrunner: I guess I'm saying that some applications may have a role for a heartbeat that cannot be implemented by KA/LWT, in that the heartbeat (not this implementation) can contain up to date information about a device.

PR #77, opened 2016-04-06
[QUARKS-121] Fix JavaDoc error except </P> tag

Cazen:


Cazen:
There's a error when building. I will attach WIP tag and finding why my commits cause some problems.

analytics/sensors/src/main/java/quarks/analytics/sensors/Filters.java
```diff
@@ -99,7 +99,7 @@ private Filters() {}
      * <LI>The first tuple.</LI>
      * </UL>
      * <BR>
-     * <img src="doc-files/deadband.png" alt="Deadband example"/>
+     * <img src="doc-files/deadband.png" alt="Deadband example">
```
dlaboss: I think the original above is what's needed.  Was the image rendering wrong for you?

Thanks for many of the other fixes.

For what it's worth, it seems only the "<" needs to be escaped to avoid confusing browsers.  ">" is OK.  So for example, the changes to Connector below seem unnecessary.  Were you seeing a rendering problem?

Also &lt;pre> and/or &lt;code> provide some "protection" as well - e.g., line 157 in TStream below already renders correctly.

FWIW, I've taken to trying to use the following idiom as it seems "stuff" can then basically anything, particularly unescaped "<", so I can just write / copy code in its natural form
&lt;pre>{@code
stuff...
}&lt;/pre>

dlaboss:
FYI, the build failure was due to an MQTT test that uses a public server.  I suspect it will be fine the next time.

Cazen:
Good Day [~dlaboss] Thank you for advising :)

1. The image rendering was good. I can see the picture properly.
But in Java8, There are strict rules when generate JavaDoc(it's not valid in HTML)
```
Filters.java:102: error: self-closing element not allowed
     * <img src="doc-files/deadband.png" alt="Deadband example"/>
       ^
```
Of course, we can just using -Xdoclint:none option to turn off, But I thought it's good to remove self-closing tag for avoid confusing.

2. I can see the '>' character in JavaDoc from Connector.java:49 of master branch correctly. 
I just change it to with escapse, because it cause JavaDoc generate error. 
```
Connector.java:49: error: bad use of '>'
 *                      -->C1
                          ^
```
But I think it's good idea to using '@code' tag rather than using escape like you mentioned. I will change it immediately.

3. <code> tag and {@code} tag are different in JavaDoc(http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html)
<code> tag does not replace "<" or ">" automatically. It cause error
```
{@code text}
Introduced in JDK 1.5

Equivalent to <code>{@literal}</code>.

Displays text in code font without interpreting the text as HTML markup or nested Javadoc tags. This enables you to use regular angle brackets (< and >) instead of the HTML entities (&lt; and &gt;) in documentation comments, such as in parameter types (<Object>), inequalities (3 < 4), or arrows (<-). For example, the documentation comment text {@code A<B>C} displayed in the generated HTML page unchanged as A<B>C. This means that the <B> is not interpreted as bold and is in code font. If you want the same functionality without the code font, then use the {@literal} tag.
```
I remained all '<' and '>' characters in {@code} tag, but replaced in <code> tag.(Like Tstream) to avoid error
I can replace them all from <code> tag to {@code} if you agree about that.

dlaboss:
Ah, thanks for the clarification on  and the other problems.
If something needs to be fixed and using {@code} works then that seems like the best solution to me.
I didn't mean to imply that you *should* hunt down and convert all uses of &lt;code>, even non-problem ones, to using {@code} though that may help avoid future problems... and I would support your efforts :-)

Cazen:
I really appreciate your review @dlaboss :)
Have a good day!

dlaboss:
The project really appreciates your contributions!
If you are going to make more changes it would be good to add [WIP] to the title of the PR until your final set of changes are all set to go / ready for review and merging.

Cazen:
Good day!
Please let me know if needed to add things or wrong in my work :)

PR #79, opened 2016-04-07
QUARKS-13: Start over Quarks when the JVM crashes (Linux)

vdogaru:
Simple mechanism to restart a Quarks application if the JVM running Quarks terminates unexpectedly:
 - Sample application which simulates a JVM crash after processing N tuples
 - Script which checks whether the Quarks JVM is still running, otherwise it restarts the Quarks sample app
 - Cron entry which runs the script every mnute
 - Readme providing setup instructions

### DIFF
--- a/scripts/cron/startapp.cron
+++ b/scripts/cron/startapp.cron
@@ -1,0 +1,27 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Crontab file which contains settings for scheduling the execution of a 
+# monitoring script every minute using cron.
+#
+# You must set the QUARKS environment variable to point to your Quarks 
+# installation:
+# QUARKS=/home/your_userid/quarks/java8
+#
+# Make sure that java is in the PATH environment used by cron:
+# PATH=/opt/my_java/bin:$PATH
+
+* * * * * $QUARKS/scripts/cron/startapp.sh /tmp >> /tmp/quarks-cron.log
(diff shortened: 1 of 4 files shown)


dlaboss:
perhaps a stupid question:  I'm missing the need for / benefit of the cron aspect of this solution.  Why not just have a start script that loops: run the app in the foreground and when it exits, simply restart it... and avoid the cron stuff?

ddebrunner:
I was thinking the same thing, but then what happens if that looping script dies?

scripts/cron/startapp.cron
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Crontab file which contains settings for scheduling the execution of a 
+# monitoring script every minute using cron.
+#
+# You must set the QUARKS environment variable to point to your Quarks
```
ddebrunner: Wouldn't be easier to have the quarks location passed in as a command line argument rather than requiring an evironment variable set elsewhere?

scripts/cron/startapp.cron
```diff
@@ -0,0 +1,27 @@
...
+#
+# Crontab file which contains settings for scheduling the execution of a 
+# monitoring script every minute using cron.
+#
+# You must set the QUARKS environment variable to point to your Quarks 
+# installation:
+# QUARKS=/home/your_userid/quarks/java8
+#
+# Make sure that java is in the PATH environment used by cron:
+# PATH=/opt/my_java/bin:$PATH
```
ddebrunner: Why not use the standard approach of JAVA_HOME?

scripts/cron/startapp.sh
```diff
@@ -0,0 +1,119 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+[ ! -n "${QUARKS:-}" ] && QUARKS=../..
+
+#
+# Application name identifies the running process.  For the scope of this
```
ddebrunner: Seems like with minor changes this could change from a sample to a reusable script, e.g. passing in the classpath and main class name.

vdogaru:
@dlaboss one can use cron to run a command at boot time without having root access, for example for starting a Quarks application.

vdogaru:
I made the following changes based on your feedback:
 - startapp.sh is reusable - it takes the class path and the name of the class to be launched as command line arguments.  It also uses JAVA_HOME, rather than expecting java to be in the path.
 - startapp.cron is a sample crontab entry for restarting sample TerminateAfterNTuples every minute

PR #81, opened 2016-04-08
[QUARKS-134] Add missing License header

Cazen:


Cazen:
I will find out which license used by sankey_quarks.js

Cazen:
- sankey_quarks.js is under BSD 3-Clause License(https://opensource.org/licenses/BSD-3-Clause). So no license header needed.
- But, in case of d3.BSD.LICENSE, sankey.d3.BSD.LICENSE, it looks like just license template.
  Could I change it in order to original d3 license?

  d3.BSD.LICENSE - https://github.com/mbostock/d3/blob/master/LICENSE
  sankey.d3.BSD.LICENSE - https://github.com/d3/d3-plugins/blob/master/LICENSE

ddebrunner:
Should we be changing headers on third party files (anything under `ext`)?

Cazen:
Good Day @ddebrunner 
- We need only attach header to source file that we've created
(http://www.apache.org/legal/src-headers.html -> SOURCE FILE HEADERS FOR CODE DEVELOPED AT THE ASF)
- But, in case of graph.js, index.js, metrics.js in ext folder, it looks like our own source code. So I've worked on it.
- D3 related js files are under BSD license. So I didn't attach Apache license headers on that(TREATMENT OF THIRD-PARTY WORKS)

Cazen:
I could have been misunderstood your question
If your question is about d3.BSD.LICENSE, sankey.d3.BSD.LICENSE, and you think that it is better to keep them all, please just tell me.
I thought it's weird that the license looks just a BSD template when I searched about BSD license usage in other Apache project(e.g https://github.com/apache/spark/blob/master/licenses/LICENSE-d3.min.js.txt)
Please don't hesitate to advise me.

ddebrunner:
Commit 1e154be8668c06e2b6d18c0eb12306f35cd23d07 is good. It is adding a licence header to Quarks code.

Commit 2fd31c76da043e53391c51c0503fd1df5df802be is not. Any file under an `ext` folder is a third-party distribution (or subset of a distribution). We should not be modifying third-party files, especially modifying their licence/notice files.

Now maybe there is an issue with what is in that `ext` directory, but that I think that would be better covered as a separate issue where a discussion of the correct fix (if any) can take place.

In general it's better to have pull requests and issues focus on a single topic, rather than combining multiple unrelated items. Merging pull requests is easy for committers so don't worry about creating many, the more focused a set of changes the more likely they will be merged quickly.

So I'm -1 on this pull request, but if 1e154be8668c06e2b6d18c0eb12306f35cd23d07  (or its changes) was submitted as a new pull request that would be merged quickly.

ddebrunner:
That's cool, if you put the full commit id in a comment, it gets auto-linked to the commit.

Cazen:
I worried how I can revert the commit 2fd31c7, decided to resolve the issue through revert.(https://github.com/blog/2019-how-to-undo-almost-anything-with-git) I think our discussion history will be lost if I rebase my branch in this commit.
I will revert wrong commit immediately. I agree it's better that PR should concentrate one issue rather than more. And license is sensitive issue, too. I will note.
Text comment that linking to commit is good function in github. That's interesting. haha

ddebrunner:
I think the safe way is to create a new branch in your fork and cherry-pick 1e154be8668c06e2b6d18c0eb12306f35cd23d07
so that the actual repo shows no changes to the files under ext.

Cazen:
Why not? I will follow your way. Thank you

PR #86, opened 2016-04-11
[QUARKS-131] expose control service for poll() period

dlaboss:
Looking for feedback for changes to the API.
Impl will follow once this is settled.

Due to initial concerns about ``TStream.alias(String)`` being too general (alias for what?, might there be other aliases later?), I named the method ``controlServiceAlias()``.  In retrospect probably ``alias()`` is fine.  It's an "alias for the stream" that might be used in multiple contexts, one of which is to get controls registered for the stream.

Note: relocated PeriodicMXBean as pure Topology/TStream users should not
have to know anything about the quarks.oplet domain.

### DIFF
--- a/api/execution/src/main/java/quarks/execution/mbeans/PeriodicMXBean.java
+++ b/api/execution/src/main/java/quarks/execution/mbeans/PeriodicMXBean.java
@@ -16,14 +16,14 @@ KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
 */
-package quarks.oplet.core.mbeans;
+package quarks.execution.mbeans;
 
 import java.util.concurrent.TimeUnit;
 
 /**
- * Control interface for a periodic oplet.
+ * Control interface for a periodic source stream.
  * 
- * @see quarks.oplet.core.PeriodicSource
+ * @see quarks.topology.Topology#poll(quarks.function.Supplier, long, TimeUnit)
  *
  */
 public interface PeriodicMXBean {
@@ -44,4 +44,9 @@ public interface PeriodicMXBean {
      * Set the period.
      */
     public void setPeriod(long period);
+    
+    /**
+     * Set the period and unit
+     */
+    public void setPeriod(long period, TimeUnit unit);
 }
--- a/api/graph/src/main/java/quarks/graph/Connector.java
+++ b/api/graph/src/main/java/quarks/graph/Connector.java
@@ -123,7 +123,8 @@ public interface Connector<T> {
      * The alias must be unique within the topology.
      * The alias may be used in various contexts:
      * <ul>
-     * <li>Runtime control services for the connector are registered with this alias.</li>
+     * <li>Runtime control services for the Connector (stream/outputport)
+     * are registered with this alias.</li>
      * </ul>
      * </p>
      * 
--- a/api/graph/src/main/java/quarks/graph/Connector.java
+++ b/api/graph/src/main/java/quarks/graph/Connector.java
@@ -115,5 +115,27 @@ public interface Connector<T> {
      * 
      * @return set of tag values.
      */
-    Set<String> getTags(); 
+    Set<String> getTags();
+    
+    /**
+     * Set the alias for the connector.
+     * <p>
+     * The alias must be unique within the topology.
+     * The alias may be used in various contexts:
+     * <ul>
+     * <li>Runtime control services for the connector are registered with this alias.</li>
+     * </ul>
+     * </p>
+     * 
+     * @param alias the alias
+     * @throws IllegalStateException if the an alias has already been set
+     */
+    void alias(String alias);
+    
+    /**
+     * Returns the alias for the connector if any.
+     * @return the alias. null if one has not be set.
+     */
+    String getAlias();
+
 }
--- a/api/oplet/src/main/java/quarks/oplet/OpletContext.java
+++ b/api/oplet/src/main/java/quarks/oplet/OpletContext.java
@@ -76,6 +76,12 @@ public interface OpletContext<I, O> extends RuntimeServices {
     List<? extends Consumer<O>> getOutputs();
 
     /**
+     * Get the oplet's output port context information.
+     * @return list of {@link OutputContext}, one for each output port.
+     */
+    List<? extends OutputContext> getOutputContext();
+
+    /**
      * Get the job hosting this oplet. 
      * @return {@link JobContext} hosting this oplet invocation.
      */
--- a/api/oplet/src/main/java/quarks/oplet/OpletContext.java
+++ b/api/oplet/src/main/java/quarks/oplet/OpletContext.java
@@ -77,9 +77,9 @@ public interface OpletContext<I, O> extends RuntimeServices {
 
     /**
      * Get the oplet's output port context information.
-     * @return list of {@link OutputContext}, one for each output port.
+     * @return list of {@link OutputPortContext}, one for each output port.
      */
-    List<? extends OutputContext> getOutputContext();
+    List<OutputPortContext> getOutputContext();
 
     /**
      * Get the job hosting this oplet. 
--- a/api/oplet/src/main/java/quarks/oplet/OutputContext.java
+++ b/api/oplet/src/main/java/quarks/oplet/OutputContext.java
@@ -16,37 +16,15 @@ KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
 */
-package quarks.execution.mbeans;
-
-import java.util.concurrent.TimeUnit;
+package quarks.oplet;
 
 /**
- * Control interface for a periodic entity.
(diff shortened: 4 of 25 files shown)


api/execution/src/main/java/quarks/execution/mbeans/PeriodicMXBean.java
```diff
@@ -16,14 +16,14 @@ Licensed to the Apache Software Foundation (ASF) under one
...
 under the License.
 */
-package quarks.oplet.core.mbeans;
+package quarks.execution.mbeans;
 
 import java.util.concurrent.TimeUnit;
 
 /**
- * Control interface for a periodic oplet.
+ * Control interface for a periodic source stream.
```
ddebrunner: Since this has moved it now also seems completely generic it could be used for periodic entity, therefore:

`s/source stream/entity/`

?

ddebrunner:
+1 

Like the move of PeriodicMXBean.

api/execution/src/main/java/quarks/execution/mbeans/PeriodicMXBean.java
dlaboss: makes sense

dlaboss:
For impl review, see prior commit's comment for design choice:  OutputContext and OpletContext

dlaboss:
@ddebrunner, @vdogaru pls review the impl as the changes are in unfamiliar territory. 
see the "conversation" stream above for some info.
thx!

api/graph/src/main/java/quarks/graph/Connector.java
```diff
@@ -115,5 +115,27 @@ Licensed to the Apache Software Foundation (ASF) under one
...
-    Set<String> getTags(); 
+    Set<String> getTags();
+    
+    /**
+     * Set the alias for the connector.
+     * <p>
+     * The alias must be unique within the topology.
+     * The alias may be used in various contexts:
+     * <ul>
+     * <li>Runtime control services for the connector are registered with this alias.</li>
```
vdogaru: I'm a bit confused by "Runtime control services for the connector".  At the high level, a control service provides an interface for controlling the stream and is registered with the TStream's alias.

api/topology/src/main/java/quarks/topology/TStream.java
```diff
@@ -443,7 +443,7 @@ else if (WARNING.equals(lr.getLevel()))
     Set<String> getTags(); 
     
     /**
-     * Add an alias for the stream.
+     * Set an alias for the stream.
      * <p>
      * The alias must be unique within the topology.
```
vdogaru: Is the statement "must be unique within the topology" too restrictive? 
The system registers a control associated with that TStream using ```ControlService.registerControl(String type, String id, String alias, Class<T> controlInterface, T control)```, where the 'alias' parameter is TStream's alias. The registration requires that 'alias' is unique within the context of the control 'type'.  So I could have:
```
TStream<Double> t1  = topology.poll(...);
t1.alias("foo");
// At run time, the PeriodicSource oplet registers a PeriodicMXBean with type "periodic" and alias "foo"
...
TStream<String> t2 = t1.map(...);
t2.alias("foo");
// At run time, the Map oplet registers a (hypothetical) MapMXBean with type "map" and alias "foo"
```

dlaboss: I think it has to be that restrictive for things to remain sane.  What you say is true, but a TStream.alias() user has no idea or control over what "types" may be registered by the runtime or others using an alias so they can't predict or otherwise avoid collisions, right?   

E.g., maybe some day poll() and map() register some mbean with the same type, say, some logging mbean or such?

At a higher level I'm not really getting it wrt picking "type" or alias names, or "ids" for that matter in order to generally avoid collisions since the ControlService is documented as expecting to be used by application code as well as the runtime.  Seems like minimally more doc is needed in ControlService for conventions for these.  And right now the runtime uses unadorned values like "job" and "periodic" for types - what's to prevent an ignorant user for picking those to use for their own application specific controls?

api/graph/src/main/java/quarks/graph/Connector.java
dlaboss: Sort of confusing since there are really 3 abstraction levels to the system :-)  Would changing "... the connector" => " the Connector" (talking about this class not connectors in the sense of mqtt, iot, ...) or "this Connector" help clarify?

[Maybe I'm just misreading your second sentence.  There's nothing stream/TStream specific about a control service nor ControlService. ]

At the API level users deal in terms of TStream.alias() - an alias for the stream and from their perspective control services are registered for the stream.

But at the graph level of the system, its a graph.Connector (the logical equivalent of an output port / stream) that has the alias and services are registered are for the Connector's alias.  It has no direct knowledge of TStream.

The oplet level, doesn't have any knowledge of TStream or Connector, it just deals in Consumer for an output port.  It needs to register the control service with the alias from the "output port" (aka stream / Connector / TStream).   I'm particularly interested in a sanity check on the addition of OutputContext to OpletContext.  It, or something else, is needed for the oplet to be able to get the alias to register mbean.

dlaboss:
See QUARKS-147 (adaptable TWindow width) for motivation for the PeriodicMXBean=>PeriodMXBean change and TStream.TYPE and change in registration.

api/graph/src/main/java/quarks/graph/Connector.java
vdogaru: > at the graph level ... services are registered for the Connector's alias
Right, Connectors live as long as the Graph, so can serve as registration object.

+1 for OutputContext.

api/topology/src/main/java/quarks/topology/TStream.java
vdogaru: Maybe we need a naming convention similar to the proposal for tag values - https://issues.apache.org/jira/browse/QUARKS-68 ?

api/oplet/src/main/java/quarks/oplet/core/PeriodicSource.java
```diff
@@ -47,8 +48,8 @@ public void initialize(OpletContext<Void, T> context) {
     public synchronized void start() {
         ControlService cs = getOpletContext().getService(ControlService.class);
         if (cs != null)
-            cs.registerControl("periodic", getOpletContext().uniquify(getClass().getSimpleName()), 
-                    getAlias(), PeriodicMXBean.class, this);
+            cs.registerControl(TStream.TYPE, getOpletContext().uniquify(getClass().getSimpleName()),
```
vdogaru: Would it make sense to register the control against a class type rather than a String?  Instead of:

`<T> String registerControl(String type, String id, String alias, Class<T> controlInterface, T control);` where 
the `type` parameter identifies the type of the control MBean, use:

`<T> String registerControl(Class<?> type, String id, String alias, Class<T> controlInterface, T control);` where the `type` parameter is the type of object controlled by the MBean.

api/oplet/src/main/java/quarks/oplet/OpletContext.java
```diff
@@ -76,6 +76,12 @@ Licensed to the Apache Software Foundation (ASF) under one
     List<? extends Consumer<O>> getOutputs();
 
     /**
+     * Get the oplet's output port context information.
+     * @return list of {@link OutputContext}, one for each output port.
+     */
+    List<? extends OutputContext> getOutputContext();
```
ddebrunner: Why `? extends OutputContext` ?

api/oplet/src/main/java/quarks/oplet/OutputContext.java
```diff
@@ -0,0 +1,30 @@
...
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/
+package quarks.oplet;
+
+/**
+ * Information about an oplet output port. 
+ */
+public interface OutputContext {
```
ddebrunner: Should this be `OutputPortContext` given its description?

dlaboss: Sure, works for me!  I initially omitted "port" because the oplet layer seems to omit it - e.g., ``OpletContext.getOutputs()``.

api/topology/src/main/java/quarks/topology/TStream.java
dlaboss: https://issues.apache.org/jira/browse/QUARKS-154 is for "type" naming conventions doc/improvements.

api/oplet/src/main/java/quarks/oplet/core/PeriodicSource.java
dlaboss: added the above idea to QUARKS-154

api/oplet/src/main/java/quarks/oplet/OpletContext.java
dlaboss: fixed

PR #93, opened 2016-04-19
[QUARKS-150] add Valve Predicate

dlaboss:
Q: quarks.analytics.sensors or quarks.topology.plumbing?

Note: there's no benefit to a Filters.valve(State initialState) because in order for the app to change the state of the valve the app needs to new the Valve itself so that it can eventually do valve.setState(newState).  That could change if Valve supports a control mbean but that's a more general discussion for functions an mbeans.


ddebrunner:
Given its name, plumbing would seem more appropriate.

analytics/sensors/src/main/java/quarks/analytics/sensors/Valve.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.Objects;
+
+import quarks.function.Predicate;
+
+/**
+ * A generic "valve" {@link Predicate}.
+ * <p>
+ * A valve predicate accepts tuples when its state is {@link State#OPEN},
+ * otherwise it rejects tuples.
```
ddebrunner: maybe discards or ignores tuples, it doesn't really 'reject' them.

analytics/sensors/src/main/java/quarks/analytics/sensors/Valve.java
```diff
@@ -0,0 +1,91 @@
...
+ * @param <T> tuple type
+ */
+public class Valve<T> implements Predicate<T> {
+    private static final long serialVersionUID = 1L;
+    private transient State state = State.OPEN;
+    
+    /**
+     * The valve state.
+     */
+    public enum State {
```
ddebrunner: Is there value to an enum as opposed to just a boolean? It seems to be extra (minimal) footprint for little benefit.

PR #96, opened 2016-04-21
QUARKS-146 [WIP]

saurabhjinturkar:
Replaced e.printStackTrace() statements with logger errors.

### DIFF
--- a/connectors/iotf/src/main/java/quarks/connectors/iotf/runtime/IotfDeviceCommands.java
+++ b/connectors/iotf/src/main/java/quarks/connectors/iotf/runtime/IotfDeviceCommands.java
@@ -45,7 +45,7 @@ public class IotfDeviceCommands implements Consumer<Consumer<Command>> {
         try {
             connector.subscribeCommands(commandSubmitter);
         } catch (Exception e) {
-			logger.error("Exception caught while subscribing commands: {}", e);
+            logger.error("Exception caught while subscribing commands", e);
         }
     }
 }
--- a/connectors/iotf/src/main/java/quarks/connectors/iotf/runtime/IotfDeviceCommands.java
+++ b/connectors/iotf/src/main/java/quarks/connectors/iotf/runtime/IotfDeviceCommands.java
@@ -19,6 +19,9 @@ under the License.
 
 package quarks.connectors.iotf.runtime;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.ibm.iotf.client.device.Command;
 
 import quarks.function.Consumer;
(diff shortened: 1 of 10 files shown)


connectors/iotf/src/main/java/quarks/connectors/iotf/runtime/IotfDeviceCommands.java
```diff
@@ -41,8 +45,7 @@ public void accept(Consumer<Command> commandSubmitter) {
         try {
             connector.subscribeCommands(commandSubmitter);
         } catch (Exception e) {
-            // TODO Auto-generated catch block
-            e.printStackTrace();
+			logger.error("Exception caught while subscribing commands: {}", e);
```
dlaboss: Coding conventions are missing from DEVELOPMENT.md but that effort is underway (e.g., QUARKS-23).
For the most part the code uses spaces instead of hard tabs, with 4 spaces of indentation.  (the above may be changing that to 2).  Please remove the hard tabs.

The ": {}" in the error() msg string needs to be removed here and in other cases below.
I noticed that the sample use that I pointed you at in JobRegistry had that problem (I notified the author).  Sorry about that.

Thanks for your help!

dlaboss:
The travis HttpServerTest failure there is fallout from adding use of slf4j to ConsoleJobService.
Presumably a one of the console's build.xml needs to be augmented to include slf4j but it wasn't obvious to me which/where as both seem to include things that pull it in.

@home4slc ,  you can make the addition to ConsoleJobService in your workspace and rebuild and run the HttpServerTest to reproduce it.  Thoughts about how to fix that?

connectors/iotf/src/main/java/quarks/connectors/iotf/runtime/IotfDeviceCommands.java
saurabhjinturkar: Thank you for the review @dlaboss. I will remove placeholder "{}" from my changes.

home4slc:
Yes @dlaboss let me try it and see if I can resolve it.

dlaboss:
The HttpServerSample mods weren't the ones inducing the test problem.  It seemed to be the mods to ConsoleJobServlet.  I *suspect* that the changes to *any* of the ``console/servlet`` files will run into this problem.  So you'll have to undo your changes to those while @home4slc investigates that.

home4slc:
I added the slf4j log files to the console build via QUARKS-157, the logging project can proceed as far as the console classes are concerned.

Thanks,

Susan

dlaboss:
I'm going to merge this now.
@saurabhjinturkar, can you start a new branch/PR and complete converting the console code now that  the fix for QUARKS-157 has been delivered?

saurabhjinturkar:
Thanks @dlaboss. I will update console part and generate a new pull request.

PR #99, opened 2016-04-26
[QUARKS-156] [FINAL-REVIEW-NEEDED] "concurrent analytics" / barrier

dlaboss:
See https://issues.apache.org/jira/browse/QUARKS-156
[ See "ready for final review" comment below for summary of things. ]

Looking for comments for the API (not the impl yet).

This proposed high level concurrentMap() is to "make the simple simple".  I suspect some lower level assets, that could also be directly used by apps, may be created to realize its implementation (more on that to come).

Questions:
- happy with the semantics and signature?
- name OK?
- should it be ``TStream.concurrentMap()`` instead?


ddebrunner:
I think the general case is that  the "concurrent operations" are a pipeline of arbitrary operations, not a single function, though then there are potential issues with missing or extra tuples.

Need to define what happens with missing tuples, i.e. a mapper returns null.

dlaboss:
Yeah, the next commit will add a 2nd "stream based" concurrent() method for feedback.  At least with this initial version, each pipeline MUST return exactly one result tuple for an input tuple.  (can offer alternatives later if needed).

I'll also update the "function based" concurrentMap() to indicate that mappers MUST return non-null.  (same "can do more later if needed").

The pipeline comment also made me think that users of the "function based" concurrentMap() would find it convenient if Quarks functions supported Java8 style function composition - e.g., ``java.util.Function.andThen()``.  Thoughts on that?  I can create a Jira to track that.

dlaboss:
I'm posting tests for the concurrentMap() and concurrent() as well as initial implementations.  The initial implementations explore what it would be like with no new oplets.  Comments on those welcome.  Real barrier is next.

dlaboss:
This is all set for final review.

Q: are things in the "best" packages?

These API enhancements were made:
- Added PlumbingStreams.concurrent()
- Added PlumbingStreams.concurrentMap()
- Added PlumbingStreams.barrier()
- Added PlumbingStreams.isolate(stream, queueCapacity) (and tweaked Isolate)
- Added TStream.fanin()
- Added quarks.oplet.core.FanIn
- Adjusted Union: isa FanIn
- Added quarks.oplet.plumbing.Barrier (isa FanIn)
- Added tests for concurrent() (implicitly barrier()/Barrier/FanIn)

TODO: add explicit test for barrier()

dlaboss:
@ddebrunner +1 on the API changes noted above?

PR #109, opened 2016-05-08
[QUARKS-166] Add Gate plumbing

Cazen:
Hmm The sample code in QUARKS-166 looks clearly. I just added test case.
Any advice would be appreciated :)

### DIFF
--- a/api/topology/src/test/java/quarks/test/topology/PlumbingTest.java
+++ b/api/topology/src/test/java/quarks/test/topology/PlumbingTest.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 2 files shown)


api/topology/src/test/java/quarks/test/topology/PlumbingTest.java
```diff
@@ -601,4 +602,66 @@ public void testParallel() throws Exception {
 //            actDuration < 0.5 * expMinSerialDuration);
 //    }
 
-}
+    @Test
+    public void testGate() throws Exception {
+        // Timing variances on shared machines can cause this test to fail
```
dlaboss: Does this "timing variances" failure condition actually happen?  I can't imagine how it could.  Same with 2nd test case (though may have to increase test-tmo beyone 1sec. for that one).

Cazen: Hi @dlaboss 
Sorry for that. At first time I was planning to make test with PlumbingStreams.blockingDelay (similar to the testBlockingDelay) However, plans changed but the line did not erased.
I will erase immediately "assumeTrue" line in two Testcase

dlaboss:
unrelated testParallelMap failed due to timing/load on ci-system.  i'll fix that in another pr.  looks like the new testsGate*() passed before encountering that.  maybe close/reopen just to give it another try.

dlaboss:
unrelated testParallelMap timing/validation failure on ci system. I'll fix that in another pr.

PR #112, opened 2016-05-09
Add the ability to change a job's state from its mbean.

ddebrunner:
and some tracing for the control & app services

### DIFF
--- a/api/execution/src/main/java/quarks/execution/mbeans/JobMXBean.java
+++ b/api/execution/src/main/java/quarks/execution/mbeans/JobMXBean.java
@@ -18,6 +18,8 @@ under the License.
 */
 package quarks.execution.mbeans;
 
+import quarks.execution.Job.Action;
+
 /**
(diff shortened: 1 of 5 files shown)


api/execution/src/main/java/quarks/execution/mbeans/JobMXBean.java
```diff
@@ -150,4 +152,13 @@ static public Health fromString(String health) {
      * @return a JSON-formatted string representing the running graph. 
      */
     String graphSnapshot();
+
```
vdogaru: I'm thinking of a JobMXBean state as a read-only attribute, reflecting changes initiated by the Etiao runtime.  I'm wondering what is the use case for initiating a job state change from a JMX client?

vdogaru: QUARKS-172: Will allow a job to be cancelled by remote agent.

PR #116, opened 2016-05-11
[QUARKS-22] StreamScope

dlaboss:
- A StreamScope (Stream Oscilloscope) captures tuples.  StreamScopes can be enabled/disabled/paused/resumed, they have count and time based captured tuple retention policy, they have count, time, or Predicate based tuple selection controls.  By default each streamScope is disabled.  A Sample is created for each captured tuple containing the tuple (not copied) and capture timestamps.  The control mbean (below) makes the captured samples available as JSON.
- The DevelopmentProvider adds StreamScope peekers to streams (akin to Metrics being added).
- StreamScope's are registered with a new StreamScopeRegistry runtime service.  The registration occurs as part of a StreamScope oplet's initialize().
- StreamScopeRegistryMXBean and StreamScopeMXBean control mbeans are defined and registered in the (development provider JMX) ControlService.  The registry mbean lazy creates / registers StreamScope mbeans.
-Tests for all of the above were added.

Notes for the Console:
- The Console will lookup a stream's StreamScopeMXBean by its streamId via the StreamScopeRegistryMXBean.  A servlet/StreamScopeUtil helper is supplied.
- A StreamScope is registered by its output stream's streamId, not the streamId of the "origin stream" it was created to monitor.  That's one of the issues alluded to below. The util class forms the streamId.
- StreamScope oplets are added to the graph (all the smarts are in the StreamScope `Consumer<T>` function).  StreamScope is a subclass of Peek.
- Console changes in this PR include:
  - present StreamScope oplets the same shape and color as CounterOp
  - propagate CounterOp metric values across peek/StreamScope oplets (so the stream segments connecting two "real" oplets looks continuous: same width & color), and to FanOut outputs
  - I suggest not presenting these metric/StreamScope instrumentation oplets at all, e.g., instead just including a "show tuples" link on a stream's hover popup.

I encountered a variety of issues while doing this work and will follow up with wiki page to enumerate and discuss them.


ddebrunner:
Seems like this should be a independent module rather than being part of the core api.

dlaboss:
@ddebrunner, does utils/quarks.streamscope and quarks.utils.streamscope.jar sound good for new location?

dlaboss:
unrelated websocket test failure

PR #121, opened 2016-05-19
[QUARKS-139] Init gradle branch

Cazen:
merging @lfeagan and @ddebrunner 's work and init branch


Cazen:
Build error seems affected by build.xml file changed. I will proceed with ignore once

PR #123, opened 2016-05-26
[QUARKS-181] Add globalization tests

queeniema:
Tests have been added for the following connectors:
* File
* HTTP
* JDBC
* Kafka
* MQTT
* PubSub
* WebSocketClient

### DIFF
--- a/connectors/file/src/test/java/quarks/test/connectors/file/FileStreamsGlobalTest.java
+++ b/connectors/file/src/test/java/quarks/test/connectors/file/FileStreamsGlobalTest.java
@@ -18,31 +18,18 @@ under the License.
 */
 package quarks.test.connectors.file;
 
-import org.junit.Test;
-
 /**
  * FileStreams connector globalization tests.
  */
 public class FileStreamsGlobalTest extends FileStreamsTest {
 
-    String[] globalLines = new String[] {
+    private static final String[] globalLines = new String[] {
             "學而時習之",
             "不亦說乎"
     };
 
-    @Test
-    public void testGlobalTextFileReader() throws Exception {
-        super.testTextFileReader(globalLines);
-    }
-
-    @Test
-    public void testGlobalTextFileReaderProblemPaths() throws Exception {
-        super.testTextFileReaderProblemPaths(globalLines);
-    }
-
-    @Test
-    public void testGlobalTextFileReaderPrePost() throws Exception {
-        super.testTextFileReaderPrePost(globalLines);
+    public String[] getLines() {
+        return globalLines;
     }
 
 }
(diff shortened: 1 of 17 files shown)


connectors/file/src/test/java/quarks/test/connectors/file/FileStreamsGlobalTest.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class FileStreamsGlobalTest extends FileStreamsTest {
+
+    String[] globalLines = new String[] {
+            "學而時習之",
+            "不亦說乎"
+    };
+
+    @Test
+    public void testGlobalTextFileReader() throws Exception {
```
dlaboss: The global strings are nicely isolated and the test code is nicely reused.  However needing to override all of the methods of the base/standard class seems unfortunate.  Can that be avoided / simplified by something along the lines of the following?
```
String[] stdLines = new String[] {  // keep the original code
  ...
}

public String[] getLines() { return stdLines; }  // override in the global test

public void testFoo() {  // orig signature
    // orig code String[] lines = stdLines;
    String[] lines = getLines();  // new code
    ... // rest of orig code
}
```

queeniema: @dlaboss, thanks for the great suggestion; I'm a bit rusty! I'll make the appropiate changes.

queeniema:
Do I need to add the new test classes to a test suite file somewhere?

ddebrunner:
Any test in a test directory with name XXXTest.java will be picked up automatically.

You should run the test suite before any commit, it does seem the new tests are failing.

queeniema:
There is one test failing: `FileStreamsTextFileWriterGlobalTest.testRetainAggSizeBased` due to: `java.lang.AssertionError:  [] expected:<2> but was:<0>`

It fails in my local testing environment as well. Something interesting: if I change the `globalStr` string to just a single character (e.g., `九` or `Θ`), I get 1 element instead of 0: `[/var/folders/sf/blbx3s152dn2mdhppv9sg3xc0000gn/T/test14407282066706266234txt_201605147_152303_3] expected:<2> but was:<1>`

If I change the `globalStr` string to `123456789` (the original string), then the test passes, which leads me to conclude that the issue is not in the test code.

Any ideas where the error might be?

dlaboss:
I'm guessing the computation of aggregateFileSize is the problem - it's computing characters while the retention "file size" is in bytes.

Try changing `long aggregateFileSize = 2 * (("1-"+getStr()).length() + 1/*eol*/);`
to `... = 2 * (("1-"+getStr()).getBytes().length + 1/*eol*/);`

queeniema:
@dlaboss That was it, thank you! Test is passing now.

PR #128, opened 2016-06-03
[QUARKS-197] tweak test's timing for intermittent failures

dlaboss:
got a couple of failures where 1,4,6,10 was received instead of 1,4,6,9.
Think the deadtime change was a bit too long so 9 wasn't outside of the
time-band.


dlaboss:
hmm... wget of ant failed.  retry.

PR #131, opened 2016-06-08
[WIP][QUARKS-139] Alternative gradle build

bjhargrave:
Here is another go at a gradle build replacement for quarks.

This build has minimal `build.gradle` files in each project and concentrates the build logic in the root project `build.gradle` file.

This build tries to use the standard gradle way of doing things. Output is placed in the standard gradle places. e.g. `build/libs` for generated jar file. Inter-project dependencies are expressed using `project(':api:topology')` form. External jars are referenced using mavenCentral to avoid the need to place binaries in the build.

The build completes and tests pass running `./gradlew build`.

This only builds the projects with Java 8. More work is of course needed to complete the build in gradle such as the Java 7 and Android part as well as any packaging steps to package the generated jars into some distribution shape.

Eventually, I want to apply the `biz.aQute.bnd.builder` plugin to generate OSGi metadata for each generated jar to allow the jars to be used in OSGi runtimes such as Eclipse Kura.


ddebrunner:
I didn't see any updates to DEVELOPMENT.md to tell folks how to build.

If this is merged to master, what's the resulting state of the build, do I need to use ant and gradle, can I continue to build to support Android, can I still build a release?

Or was this meant to be a pull request to a gradle branch?

build.gradle
```diff
@@ -0,0 +1,96 @@
...
+    testLogging {
+      exceptionFormat 'full'
+    }
+  }
+
+  ext.jarOptions = {
+    manifest {
+      attributes(
+        'Implementation-Title': "${-> baseName}",
+        'Implementation-Vendor': 'quarks-edge@github.com',
```
dlaboss: Should be apache/incubator-quarks@github.com or such I think?   quarks-edge is the decomissioned pre-ASF project.

bjhargrave:
This PR is initially for discussion about how to proceed. As I mentioned more work is needed, that I can purse, if we agree to go this route. That work will of course need to include updated to docs like DEVELOPMENT.md  and also .travis.yml.

This PR is not against the gradle branch. I started fresh from master as I felt that if we move to a gradle build, we should do things the gradle way (e.g. jars in build/libs) rather than just replace ant with gradle and continue to use the ant builds places for class file, jar files, etc. One of the points of having a gradle build is that gradle knowledgeable engineers can understand the build quickly. Having non-standard places for outputs go against that. 

This branch can be merged to master with no affect on the ant build since this gradle build puts it output in gradle standard places which are different than the places the ant build uses. So we could merge this into master and then continue to improve the gradle build while the ant build continues to function. Then when the gradle build is at functional parity, we can remove the ant build.

build.gradle
bjhargrave: Yeah, I just got that from [master](https://github.com/apache/incubator-quarks/blob/master/common-build.xml#L169). I'll amend the commit.

ddebrunner:
@bjhargrave FYI - the convention is to prefix the pull request with `[WIP]` to indicate work-in-progress

bjhargrave:
> @bjhargrave FYI - the convention is to prefix the pull request with [WIP] to indicate work-in-progress

Fixed. Thanks.

build.gradle
ddebrunner: Should be the Apache Software Foundation

dlaboss:
This is awesome!  I'm looking it over/learning and giving it a shot.

console/server/src/main/java/quarks/console/server/ServerUtil.java
```diff
@@ -45,31 +53,35 @@ private String getPath() {
...
-    };
     /**
      * Returns the File object representing the "webapps" directory
      * @return a File object or null if the "webapps" directory is not found
      */
     private File getWarFilePath() {
-        File[] foundFiles = getTopDirFilePath().listFiles(fileNameFilter);
-        if (foundFiles.length == 1) {
-            return foundFiles[0];
+        List<File> foundFiles = new ArrayList<>();
```
ddebrunner: Seems like this is a merge error, I'm assuming this file should not change because of gradle.

bjhargrave: It had to change because this code is badly dependent upon the structure of the build. It assumed that the jars all sit in target/java... getTopDirFilePath assume ../../.. from the jar to find the webapp. Since the gradle build uses normal gradle output locations, the webapp folder is elsewhere. c52ecf7ca8ec1f31ed99c4e944f02185f56833c4 is the commit to be more flexible in locating the webapps folder in the build.

ddebrunner:
The build files look good, simple and clear to understand.

build.gradle
bjhargrave: Changed to Apache Software Foundation.

console/servlets/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+plugins.apply 'war'
+
+dependencies {
+  providedCompile 'javax.servlet:javax.servlet-api:3.1.0'
+  providedCompile project(':utils:streamscope')
+  providedCompile ext_classpath
+}
+
+war {
```
Cazen: Does all the test case passed? In my case, build failed due to console.war location
Could you tell me if there is something forgotten setting that I missed?

Cazen:
Nice work! It looks very cleanly. I learned a lot in your work!

Could I kindly ask some questions?

1. I agree with your idea that follow the gradle way, such as build/libs. 
But for make apache distribution(package of jars), I think we need function that copy all the jars to specific directory. How do you think about the need for these function?

2. Unlike the Ant Build, the higher-level jar is generated. Could I delete them?
(e.g. ant only have quarks.analytics.math3.jar, quarks.analytics.sensors.jar
However, gradle has a additional empty incubator-quarks.analytics.jar)

And I think there are additional work needed as follow:
(These task is just a simple & repeated work. I can append it)
1. All gradle script should have apache license header
2. Javadoc process is missing. 
3. Update user manual(such as DEVELOPMENT.md)

console/servlets/build.gradle
bjhargrave: Yes, all the tests passed for me. The change in 5a5014f is needed to find the console.war file built in the gradle build. Make sure you have the complete branch checked out.

bjhargrave:
> Nice work! It looks very cleanly. I learned a lot in your work!

Thanks!

> 
> Could I kindly ask some questions?
> 
>     I agree with your idea that follow the gradle way, such as build/libs.
>     But for make apache distribution(package of jars), I think we need function that copy all the jars to specific directory. How do you think about the need for these function?

Yes, I agree  that we will need some task/tasks to package the built jars into a distribution. I wanted to get agreement on this current work before proceeding further.

> 
>     Unlike the Ant Build, the higher-level jar is generated. Could I delete them?
>     (e.g. ant only have quarks.analytics.math3.jar, quarks.analytics.sensors.jar
>     However, gradle has a additional empty incubator-quarks.analytics.jar)

Let me look into this.

> 
> And I think there are additional work needed as follow:
> (These task is just a simple & repeated work. I can append it)
> 1. All gradle script should have apache license header
> 2. Javadoc process is missing.
> 3. Update user manual(such as DEVELOPMENT.md)

Yes, this is further work to do.

console/server/src/main/java/quarks/console/server/ServerUtil.java
```diff
@@ -19,10 +19,18 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 package quarks.console.server;
 
+import java.io.IOException;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.net.URL;
+import java.nio.file.Files;
```
dlaboss: pls remove above FilenameFilter import as it's now no longer used and causes compile warning in Eclipse.

bjhargrave: Good catch. Thanks. I amended the commit to remove the old import.

console/server/src/main/java/quarks/console/server/ServerUtil.java
```diff
@@ -45,31 +53,35 @@ private String getPath() {
      * @return a File object
      */
     private File getTopDirFilePath() {
+        String topDirProp = System.getProperty("top.dir.file.path");
```
dlaboss: ~nit: can we use a property name that's scoped by "quarks"?  It seems this new property is required/used only to make things work in the build/test context.  If that's true then perhaps also have the name capture that: "quarks.test. ..."?

bjhargrave: That is probably a good idea. Do we want to wait until a new name is settled upon? Or just plan to change the name from quarks later?

dlaboss:
All built fine and tests passed for me. Ditto for ant build/test in the same repo.  Ditto from Eclipse. So, it seems like this can be merged w/o any ill effects to the existing mechanisms.

BJ, your thoughts re merging this now vs after more work?  Now seems good to me. Maybe capture the known TODO items in the jira so we an track it all?

I was looking at the gradle generated test reports (.../build/reports/tests/index.html).  Can a top-level summary page be generated?  OK to put that as one of the TODOs :-)

console/server/src/main/java/quarks/console/server/ServerUtil.java
dlaboss: I think now is best.  It won't make converting any harder / will just get adjusted when we handle all other occurrences.

ddebrunner:
@dlaboss but every travis ci tests has failed?

dlaboss:
Looking at the log many many tests passed before the HttpGlobalTest failed.  Probably intermittent glitch with http://httpbin.org

But yes we do need a clean run to be sure.  Not sure why it one was forced to look at the raw log this time.

BJ, to rerun the travis ci tests, either add another commit to the PR or simply close and reopen it.

bjhargrave:
> BJ, to rerun the travis ci tests, either add another commit to the PR or simply close and reopen it.

I amended a commit and pushed which will restart the travis build. Some of the test seem a bit flaky. I have another branch which changes the travis file to run the gradle build and it seems to fail sometimes in some tests which don't fail on my machine. I'll need to look more into those tests. The httpbin.org tests are among them. :-(

bjhargrave:
Success! The travis build is green :-)

> So, it seems like this can be merged w/o any ill effects to the existing mechanisms.

Yes, I think we can merge this and then I can work on additional PRs to continue to complete the gradle build.

> Maybe capture the known TODO items in the jira so we an track it all?
> 
> I was looking at the gradle generated test reports (.../build/reports/tests/index.html). Can a top-level summary page be generated? OK to put that as one of the TODOs :-)

Yes, opening some issues in Jira to track the remaining work would be a good plan.

dlaboss:
Let's use QUARKS-139 as a top-level task.  I'll take ownership because I can't assign it to you.  I'll edit the description to contain TODOs.  Please augment it as needed.  Maybe individual jiras (with "build" tag) can be created on-demand as folks attack the items.

Cazen:
@dlaboss Sure. I deleted gradle branch :) 
Thank you for review!

PR #133, opened 2016-06-09
synchronized lists used in concurrent code in tests

michaeldye:
Operations on unsynchronized lists used concurrently can fail intermittently. Tests adjusted to prevent such failures.

### DIFF
--- a/api/window/src/test/java/quarks/test/window/WindowTest.java
+++ b/api/window/src/test/java/quarks/test/window/WindowTest.java
@@ -266,8 +266,8 @@ public class WindowTest {
 		// Timing variances on shared machines can cause this test to fail
     	assumeTrue(!Boolean.getBoolean("quarks.build.ci"));
 
-        List<Long> diffs = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


wmarshall484:
+1

dlaboss:
pls close/reopen to return travis-ci (get of ant failed)

michaeldye:
Closing / reopening as requesting.

dlaboss:
looks like infr is now working.  close/reopen to rerun travis-ci.

ddebrunner:
Are we sure this is correct, and that it's not an error in the window library which should be guaranteeing synchronization?

ddebrunner:
The code in timeBatchEnsureUnique() around line 419 seems to still have synchronization issues in that it could complete before 1000 tuples have been inserted.

ddebrunner:
Ok - see that making batches synchronized in timeBatchEnsureUnique ensures that the thread running the junit test has visibility for changes made during window processing.

Another approach that would also test the window api would be to synchronize on the partition instance when performing the verification against batches, but as coded is probably how a user would use it.

PR #137, opened 2016-06-13
Add IntelliJ Idea file extension to gitignore

Cazen:
I mainly use the IntelliJ Idea, so I think it will be helpful if gitignore including an IntelliJ file extensions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,5 @@ hs_err_pid*
 build/
 
 # IntelliJ Idea
-.idea
+.idea/
 *.iml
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -18,3 +18,7 @@ hs_err_pid*
 
 /.gradle/
 build/
+
+# IntelliJ Idea
+.idea
```
bjhargrave: `.idea` is a folder, so this should end in `/`: `.idea/`

Cazen: Thank you @bjhargrave for review! I missed it.
I will modify that immediately

PR #138, opened 2016-06-14
[QUARKS-204] Release target, creating the target/… structure

Cazen:
I compared the Ant build ```./ant -> target/java8``` and Gradle build ```./gradlew build -x test -> build/distributions/java8/``` using the ```find . | grep -v ext | sort``` script.

Although some differences are found, it has a substantially similar structure

 1. Some javax related file name has been changed
- [x] javax.websocket-client.jar(ant) -> quarks.connectors.javax.websocket-client.jar(gradle)
- [x] javax.websocket-server.jar(ant) -> quarks.connectors.javax.websocket-server.jar(gradle)
- [x] quarks.javax.websocket.jar(ant) -> ./connectors/quarks.javax.websocket/lib/(gradle)

 2. There are no svt jar(quarks.test.svt.jar) exist in ant build(I think it's ant build script problem)
- [x] ./test/svt/lib/quarks.test.svt.jar

Because I already started this PR for compare directories&files schema with ant build, it has not been reflected yet about gathered all the jar files in one place with @bjhargrave advise.

### DIFF
--- a/connectors/javax.websocket-client/build.gradle
+++ b/connectors/javax.websocket-client/build.gradle
@@ -11,6 +11,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+archivesBaseName = project.name
+
 dependencies {
   compile project(':connectors:quarks.javax.websocket')
   compile 'org.eclipse.jetty.websocket:javax-websocket-client-impl:9.3.6.v20151106'
--- a/connectors/javax.websocket-client/build.gradle
+++ b/connectors/javax.websocket-client/build.gradle
@@ -11,8 +11,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-archivesBaseName = project.name
-
 dependencies {
   compile project(':connectors:quarks.javax.websocket')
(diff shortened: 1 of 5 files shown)


connectors/javax.websocket-client/build.gradle
```diff
@@ -11,8 +11,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-archivesBaseName = project.name
```
Cazen: Good day! @bjhargrave 
Could I kindly ask this property(archivesBaseName) is used?

Cazen:
image

Cazen:
I waited a while but thought it's better to proceed now and continue to gradually improving.

My base idea at participating in this issue is create a gradle build that equivalent of an ant build. I believe this way will be able to easily apply to the other nice idea(collect all of the jars to one place or some substitution of ant build's part)

I was sure that all of schema and file name to match.

dlaboss:
I think the wsclient/websockets jar differences are OK.
I think the fvt/svt jar differences are OK.
I agree that working on the console.war and other work to consolidate jar location can be done after this is merged.

PR #146, opened 2016-06-21
Fixed time-based batch windowing issue

wmarshall484:


api/window/src/main/java/quarks/window/Policies.java
```diff
@@ -69,11 +71,11 @@ Licensed to the Apache Software Foundation (ASF) under one
         
         // Can't use lambda since state is required
         return new BiConsumer<Partition<T,K,L>, T>() {
-            private boolean first = true;
+            private Set<Partition<T,K,L>> initialized_partitions = new HashSet<>();
```
dlaboss: needs to be a synchronized set

wmarshall484: Fixed in latest commit.

PR #147, opened 2016-06-22
[QUARKS-205] Create gradle Javadoc task

Cazen:
WIP - All javadoc files are generated but something phrase are different when running diff. I will check it soon

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -175,14 +175,14 @@ task aggregateJavadoc(type: Javadoc) {
     overview "quarks_overview.html"
     windowTitle "Quarks v${build_version}"
 
-    group("Quarks API", "quarks.execution", "quarks.function", "quarks.topology", "quarks.topology.*", "quarks.execution.*")
+    group("Quarks API", "quarks.execution", "quarks.function", "quarks.topology", "quarks.topology.json", "quarks.topology.mbeans", "quarks.topology.plumbing", "quarks.topology.services", "quarks.execution.*")
     group("Quarks Providers", "quarks.providers.*")
     group("Quarks Connectors", "quarks.connectors.*")
     group("Quarks Samples", "quarks.samples.*")
     group("Quarks Analytics", "quarks.analytics.*")
     group("Quarks Utilities", "quarks.metrics", "quarks.metrics.*", "quarks.streamscope", "quarks.streamscope.*")
     group("Quarks Low-Level API", "quarks.graph", "quarks.graph.*", "quarks.oplet", "quarks.oplet.*", "quarks.window")
-    group("Quarks SPI", "quarks.*.spi.*")
+    group("Quarks SPI", "quarks.topology.spi", "quarks.topology.spi.*")
   }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -28,6 +28,14 @@ ext.ext_classpath = ['com.google.code.gson:gson:2.2.4',
   'io.dropwizard.metrics:metrics-core:3.1.2']
 ext.target_dir = "$distsDir"
 ext.target_java8_dir = "${ext.target_dir}/java8"
+ext.target_javadoc_dir = "${ext.target_dir}/docs/javadoc"
+
+if (JavaVersion.current().isJava8Compatible()) {
+  tasks.withType(Javadoc) {
+    options.addStringOption('Xdoclint:none', '-quiet')
```
dlaboss: Is Xdoclint:none still required?  I did a lot of work cleaning up things but maybe I didn't get everything.  Were there many complaints?  If possible/sensible, it would seem better to just fix them.  If that's a good idea but we really don't want to tackle that at this instant, please create another jira to track doing that after we finally convert to gradle.

Cazen: Hi @dlaboss Thank you for review!
It looks only few warning in javadoc. I removed Xdoclint:none option.
I will create another jira issue for remove these warnings

And, except for a few differences, the results of Javadoc of ant and gradle are almost matches.(compare with ```diff -r build/distributions/docs/javadoc/ target/docs/javadoc/```)
The only thing I worry about is we could not see the JavaDoc that inherited from interface
(e.g ```diff -r build/distributions/docs/javadoc/quarks/connectors/file/FileWriterPolicy.html target/docs/javadoc/quarks/connectors/file/FileWriterPolicy.html``` 
And the ant output added description from interface like ```Description copied from interface: quarks.connectors.file.runtime.IFileWriterPolicy``` but gradle is not)

Unfortunately, I still have not found a way to solve this problem. I am still looking for a solution, but could you tell me if you know the related javadoc options?

dlaboss: Sorry, I don't know why the behavior might be different / why it was working in the ant build.

Unrelated to this particular problem but something to keep in mind: eventually we'll want to address QUARKS-152.  Don't know if your current work / changes will help.

Cazen:
Sorry for late. I will keep in mind your mention about QUARKS-152 :)
But due to business trip, I'll start again this issue in this thursday

Cazen:
- I could not find a function that corresponds to ```excludepackagenames``` of ant-javadoc.
I tried to using ```exclude``` to remove the runtime packages, but the results did not inherit the javadoc of interface.
```excludepackagenames``` removes the selected packages in the final output.(See [Javadoc Task](https://ant.apache.org/manual/Tasks/javadoc.html))

- So I could not be applied to ```excludepackagenames``` feature. As a result, runtime packages are included in the javadoc, ant this will be removed later.

- Some error occurs when there is no Xdoclint:none option. I added Xdoclint:none again and will register a new jira issue for resolve these errors.

PR #153, opened 2016-06-28
[QUARKS-215] Job control gets unregistered from ControlService after job gets closed

vdogaru:
WIP waiting for feedback on [Add ControlService.getControId](https://github.com/apache/incubator-quarks/pull/153/commits/02e1698e02da7ea5169295e5d4e3598836f1bc94)

### DIFF
--- a/api/execution/src/main/java/quarks/execution/services/ControlService.java
+++ b/api/execution/src/main/java/quarks/execution/services/ControlService.java
@@ -106,4 +106,18 @@ public interface ControlService {
      * @return Control Mbean or null if a matching MBean is not registered.
      */
     <T> T getControl(String type, String alias, Class<T> controlInterface);
+
+    /**
+     * Return the unique identifier for a control Mbean registered with 
+     * this service.
+     * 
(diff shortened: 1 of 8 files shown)


vdogaru:
Client code needs the control Id for example to unregister controls of closed jobs.  Retrieving the id from the control's implementation works for JSON controls, by forcing a cast from the control's MBean interface to its implementation class. It does not work for JMX controls where beans are, I believe, dynamic proxies to the implementation.  So the considered options were:
- ControlService.getControlId() - get the control id given its name, type, and interface.  Use the id to unregister the control.
- JobMXBean.getControlId() - didn't seem the right choice as the control Id is unrelated to the bean, beanc may not be necessarily registered with a ControlService, etc.

@ddebrunner , @dlaboss - any thoughts?

vdogaru:
The Travis CI build failed with error: '[junit] Testcase: keyedTimeBatchWindowTest(quarks.test.window.WindowTest):	Caused an ERROR' but tests passed fine locally.

PR #154, opened 2016-06-28
[WIP] Quarks 176 - Update Javadoc for MqttDevice.java

saurabhjinturkar:


connectors/mqtt/src/main/java/quarks/connectors/mqtt/iot/MqttDevice.java
```diff
@@ -45,7 +45,21 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+ * <pre>
+ * {
+ *     "command" : "<<Command is extracted from command topic provided in config file>>",
+ *     "tsms" : "<<Current time stamp in milliseconds>>",
+ *     "format" : "<<json or string provided in command topic>>",
+ *     "payload" : "<<Actual payload as JsonObject or String>>"
+ * }
+ * </pre>
+ * <br><br>
+ * </p>
```
saurabhjinturkar: I am new to Java streams and lambdas; after many iterations of code reading and attempts to understand the workflow, I have following questions. I am sorry if they are trivial.
 
1. The constructor throws IllegalArgumentException (Line 165 & 168) for both event and command topics if they do not end with _/fmt/json_. As per my understanding, this will restrict payload format as JSON. We are checking if topic to decide payload format. Right?
2. Do we have any sample config file which I can refer and try running this code?
3. Events method requires eventId and qos as parameters, which then calls events method and pass functions to calculate eventId, payload and qos (Line 224-239). However, eventId and qos are independent of JsonObject for payload. That means, JsonObject is published as it is. Is my understanding correct? If yes, in that case, events method does not require payload format specification.

dlaboss: so sorry for the delay.
1. Correct.  The topic suffix defines the required payload fmt and MqttDevice currently only supports /fmt/json.
2. See scripts/apps/sensorAnalytics/README.  You can edit the sensorAnalytics.properties file there to tell the sample to use a public MQTT server as noted there.
3. Correct, that overload of events() specifies tuple independent eventId and qos values and serializes the tuple to JSON and sends a MQTT message with that JSON as its payload.

The fact that a JsonObject is both a stream tuple and payload in different contexts adds to the confusion imo :-/  Let me see if I can clarify things by more clearly separating the protocol from the API.

The MqttDevice's MQTT protocol is defined by the "command" and "eventId" MQTT topics and MQTT message payload content. 

MqttDevice defines default MQTT topic formats for eventId messages and command messages but an organization / service / interoperating group of applications are free to define their own topic formats subject to the minimum requirements described in MqttDevice's javadoc.

An "event" is a MQTT message whose payload is JSON, published to an "eventId" topic.  The definition of a particular event id's JSON contents is entirely up to the application; the JSON content is opaque to MqttDevice.

Depending on which MqttDevice.events() method is used, the JSON is either the serialized TStream<JsonObject> tuple, or the serialized JsonObject that the application generates from the TStream<JsonObject> tuple.

A "command" is a MQTT message whose payload is JSON, published to a "command" topic.  The definition of a particular command's JSON contents is entirely up to the application; the JSON content is opaque to MqttDevice.

MqttDevice packages a received "command" MQTT message into a TStream<JsonObject> tuple whose contents are defined by IotDevice.command().  The tuple's CMD_FORMAT is "json" and its CMD_PAYLOAD is a JsonObject generated from the MQTT message's JSON payload.

Hope that helps :-)

dlaboss:
Please close this PR. It's superseded by https://github.com/apache/incubator-edgent/pull/294

PR #163, opened 2016-07-12
QUARKS-226 add Command connector

dlaboss:


dlaboss:
unrelated public mqtt ssh failure

PR #165, opened 2016-07-13
QUARKS-227 disable MqttOpenTest.testSsl()

dlaboss:


dlaboss:
Made it past MqttOpenTest only to encounter an Iot test failure [QUARKS-228]
Merging this one.

PR #167, opened 2016-07-15
[WIP] [COMMENTS?] [QUARKS-230] Add timer triggered window aggregations

dlaboss:


api/topology/src/main/java/quarks/topology/TWindow.java
```diff
@@ -105,8 +129,85 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+     * @param <U> Tuple type
+     * @param period how often to invoke the aggregator
+     * @param unit TimeUnit for {@code period}
+     * @param aggregator
+     *            Logic to aggregation a partition.
+     * @return A stream that contains the latest aggregations of partitions in this window.
+     * 
+     * @see #aggregate(BiFunction)
+     */
+    <U> TStream<U> timedAggregate(long period, TimeUnit unit, BiFunction<List<T>, K, U> aggregator);
```
ddebrunner: One of the issues I had with this in a previous system was that with many partitions the behaviour was not desired in that if the partition did not change the window still fired, thus wasting cpu cycles to produce the same result. Thus I wonder if it should be more along the lines of:

Aggregation of window partition on any partition change with a minimum period of `period` between aggregations.

api/topology/src/main/java/quarks/topology/TWindow.java
```diff
@@ -105,8 +129,85 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+     * This method makes it easier to create aggregation streams
+     * using the lower level {@code Window} API to construct and
+     * supply windows with configurations not directly exposed by {@code TWindow}.
+     * 
+     * @param window the window implementation
+     * @param aggregator the aggregation function
+     *  
+     * @return A stream that contains the latest aggregations of partitions in this window.
+     */
+    <U, L extends List<T>> TStream<U> process(Window<T,K,L> window, BiFunction<List<T>, K, U> aggregator);
```
ddebrunner: Can this be explained a bit more, how does the passed in window relate to the window this is defined on?

api/topology/src/main/java/quarks/topology/TWindow.java
```diff
@@ -105,8 +129,85 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+     * @param <U> Tuple type
+     * @param period how often to invoke the batcher
+     * @param unit TimeUnit for {@code period}
+     * @param batcher
+     *            Logic to aggregation a partition.
+     * @return A stream that contains the latest aggregations of partitions in this window.
+     * 
+     * @see #batch(BiFunction)
+     */
+    <U> TStream<U> timedBatch(long period, TimeUnit unit, BiFunction<List<T>, K, U> batcher);
```
ddebrunner: Not sure I understand this. Batch processing is process all the tuples as a batch and then discard them, how would it be different to the timedAggregate?

dlaboss: a timedAggregate() aggregation invocation doesn't evict anything afterwards (only the window's count/time config affects eviction, just like the non-timed aggregate()).  Hence a tuple, if it isn't evicted between aggregator invocations, will be included in multiple aggregations.  A batched aggregation, timer triggered or content change triggered, always evicts everything hence a tuple will be included in at most one aggregation.

api/topology/src/main/java/quarks/topology/TWindow.java
dlaboss: Uh... :-)  Yeah, something's not quite right in that the TWindow.last(...) established the time-based/count-based nature of the TWindow and as presented above, the supplied Window can violate that.  Maybe it's more like TStream.last(window, processor)?  Mostly I was just trying to expose a way the make it easier to utilize the Window API without having to know about other impl details like synchronizing the processor, creating an Aggregator oplet, ...

api/topology/src/main/java/quarks/topology/TWindow.java
ddebrunner: Still a bit lost, `batch` against `last(N)` or `last(time)`, processes the tuples once in batch mode.

What does this do that is different, e.g. how does this behave against a last(3) or last(10, SECONDS) ?

api/topology/src/main/java/quarks/topology/TWindow.java
dlaboss: A reasonable question.  
In the particular use case that came up, it was OK/desirable that an unchanged partition still yielded an aggregation.  e.g., the, less than perfect, interface that was desired between the device and the iothub was to publish events on the "current location" of the device even if it hadn't moved a meaningful distance.
A more efficient, less chatty, device/iothub interface would have been to only publish under that condition.
So maybe a timed-aggregator interface that only supported timed-trigger-if-changed semantics might not be acceptible?

ddebrunner: I was wondering if the continue sending updates even if nothing has changed might be better handled by a separate operation, then it could be applied to anything, rather than just a window.

Something like pass any input tuple to the output, but send the last tuple if nothing has been received for the declared period.

dlaboss: ... per-partition.  Sounds like that separate operation could be a count(1)-timedAggregate-evenIfDidntChange.  Does it make sense to force such a user to have to use a timed-trigger-if-changed window followed by this separate operation rather than just use a count(N)-timedAggregate-evenIfDidntChange? (I can imagine it would be ok, just want to be sure)

api/topology/src/main/java/quarks/topology/TWindow.java
dlaboss: So you're asking "why have timedBatch", right?

A timedBatch and timedAggregate yield the same aggregations when the trigger interval is => the effectiveWindowWidth... because the effectiveWindowWidth will have flushed all tuples since the last batch/agg in both cases.  "effectiveWindowWidth" => with predictable/regular tuple arrival rates, one can specify a last(N) and last(S sec) that yield the same window content at any time.  e.g., with a 1 tup/sec arrival rate, last(10) and last(10sec) are equivalent.

The behavior of the two is different if the trigger rate is shorter than the effectiveWindowWidth.
e.g., last(10) with a 1 tup/sec arrival rate
timedBatch(3sec) - agg1[1-3], agg2[4-6], agg3[7-9],agg4[10-12],agg5[13-15]
timedAgg(3sec) - agg1[1-3], agg2[1-6], agg3[1-9],agg4[3-12],agg5[6-15]
Right?

That said, observe that the last(10)-timedBatch(3sec) with 1tps arrival yields the same result as either last(3)-batch() or last(3sec)-batch().  Right?

So if timedBatch yields the same result as timedAggregate when the trigger period is >= effectiveWindowWidth, and it yields the same result as those untimed batch() when timedBatch trigger period is < effectiveWindowWidth, then why have timedBatch()?

I believe there are only equivalences in the cases where the tuple arrival is regular/reliable -- not bursty or lossy.  In other cases I don't think you can come up with a last(N) and last(Ssec) that are equivalent.  Hence none of these equivalences are possible.  Hence timedBatch() isn't redundant.  e g., even the trigger period >= effectiveWindowWidth case, when there is burstyness, a last(10sec) window can contain different collections of tuples than a last(N) window, for any N (more or less depending on the burstyness and values of N).  Right?

Am I missing/misthinking something and there are always equiv configs to any timedBatch() config?  Where's a signal-processing guy when you need them? :-)

ddebrunner:
On timed batch:

I think part of my concern is that the api is meant to separate out what defines the window contents (TWindow) from how it is processed (methods on TWindow). 

timedBatch seems to be mixing the concepts of what is contained in the window and when it is processed, by say evicting tuples in last(10) window after a second, so that the contents of the window no longer match its definition.

It seems that it's really trying to define a window like:

`last(long period, TimeUnit unit, int max)`

where the contents of the window is the last `period` seconds with a maximum of the `max` most recent tuples.

Then batch is applied to this window as normal.

ddebrunner:
Maybe a better question for `timedBatch(3, SECONDS, func)`, is when would it be used over

`last(3, SECONDS).batch(func)`

?

PR #171, opened 2016-07-23
[QUARKS-223] Junit parity

Cazen:
- I'm working on but little more testing needed
- I did not want to call the ant.jar in gradle but was unable to add the file to the META-INF/service. Please teach me if anyone has a good idea.
- I modified ServerUtil.java to avoid error when there are 2 or more console.war(in webapp directory) but does not looks good. I think this remainning in a temporarily until @dlaboss 's work

### DIFF
--- a/api/topology/build.gradle
+++ b/api/topology/build.gradle
@@ -18,12 +18,16 @@ dependencies {
   compile ext_classpath
 }
 
-
-jar {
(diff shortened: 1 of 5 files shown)


Cazen:
I checked it works in Linux/OSX include all junit testcase

PR #172, opened 2016-07-26
Increased tolerance for keyedTimeBatchWindowTest. Fixed race conditions.

wmarshall484:


dlaboss:
seems like this PR can be tied to QUARKS-219 and it can be marked inprogress.

PR #181, opened 2016-08-18
Enable Edgent to publish to maven repository

xuanyue202:
WIth the updated build.gradle, one can publish edgent to local maven repository via:

gradle publishToMavenLocal

This is especially useful when edgent is used with another Maven/Gradle project.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,10 +31,10 @@ System.out.println("version=${version}")
 group = 'org.apache.edgent'
 
 def getDate() {
-        if (!hasProperty('buildDate') || buildDate == null) {
-                ext.buildDate = new java.util.Date().format('yyyyMMddHHmm');
-        }
-        return ext.buildDate;
+  if (!hasProperty('buildDate') || buildDate == null) {
+     ext.buildDate = new java.util.Date().format('yyyyMMddHHmm');
+  }
+  return ext.buildDate;
 }
 
 getDate();
@@ -150,9 +150,9 @@ subprojects {
   }
 
   task sourceJar(type: Jar) {
-                // baseName-appendix-version-classifier.extension
-                from sourceSets.main.allJava
-                classifier = 'sources'
+    // baseName-appendix-version-classifier.extension
+    from sourceSets.main.allJava
+    classifier = 'sources'
   }
 
 
@@ -183,19 +183,17 @@ subprojects {
   }
 
   publishing {
-                publications {
-                        mavenJava(MavenPublication) {
-                                from components.java
-                                artifactId = artifact
-                                artifact sourceJar
-                                //artifact javadocJar
-                                //artifact testJar
-                                pom.withXml {
-                                        // Example of appending more information to pom.
-                                        //asNode().appendNode('description', 'A demonstration of Maven POM customization')
-                                }
-                        }
-                }
+     publications {
+        mavenJava(MavenPublication) {
+           from components.java
+           artifactId = artifact
+           artifact sourceJar
+           //artifact javadocJar
+           //artifact testJar
+           pom.withXml {
+           }
+        }
+     }
   }
 
   copyJar.dependsOn assemble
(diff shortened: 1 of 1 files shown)


mvherweg:
I like this. Wanted this so badly that i wrote Maven pom's for every piece to build the project with. :(

build.gradle
```diff
@@ -29,6 +29,21 @@ allprojects {
 
 apply plugin: 'java'
 
+/* SET PROJECT INFO for all artifacts */
+assert hasProperty('version')   // See version in gradle.properties.
+System.out.println("version=${version}")
```
dlaboss: First let me say Thanks for the Contribution!  I'm new to gradle so excuse nubie questions.

I don't understand the value of this assert & println.
All projects have a 'version' property so the assert will never fail.
The println prints `version=unspecified` because the property hasn't been set -- gradle.properties defines `build_version` (later in the script it sets a subproject's version to build_version  -- I don't know why the author chose to do it that way).
Should the above lines just be removed / were they for your initial debugging?

Regarding 'group', it feels to me like it should be handled to version / build_version, originating from gradle.properties.  Your thought on that?

build.gradle
```diff
@@ -29,6 +29,21 @@ allprojects {
...
+group = 'org.apache.edgent'
+
+def getDate() {
```
dlaboss: I don't understand why getData() was added / how it's used.  Do other things use the ext.buildDate property that it establishes, or do other things invoke getDate() if its defined?

xuanyue202: This getDate function counts on the build system to append date to the build version. I think you can safely move it without side effects

build.gradle
xuanyue202: You are right, I did not take care of version well. Should do it using the build_version. 
Regarding to group, there is another question. I think the group for edgent is a little weird. It seems to me that a better setting is to unify all groups to org.apache.edgent instead of having different group ids across the project. But that may involve some additional work, plus I am not sure if there are special reasons for quarks to do so, I just leave it as is.

xuanyue202:
I made an update on my repo to 1) delete the use of getDate 2) use build_version instead of a separate variable 3) change the group id and artifacts id (using the unified groupid and move the diff part to artifacts) eg. edgent.apps:iot  => org.apache.edgent:apps.iot. This solves the artifacts name conflicts like there are multiple "graph" and "topology" in api & spi.

build.gradle
```diff
@@ -206,6 +217,19 @@ subprojects {
...
+        mavenJava(MavenPublication) {
+	   groupId "org.apache.edgent"
+           artifactId "$project.group".replace("edgent.", "") + "." + artifact
+	   version build_version
+           artifact sourceJar
+
+	   from components.java
+        }
+     }
+  }
```
dlaboss: The other day, I was playing with this, came to appreciate your comment about the unique groups, and I tried doing essentially the same thing :-)
I thought specifiers like  `org.apache.edgent:edgent.api.topology:0.4.0` would be better
I had migrated setting the project's version to allprojects ( `project.version = build_version`) so I didn't need the version spec as it defaults to the project version.
And I added `build_group: org.apache.edgent` to `gradle.properties` just to keep those identification strings in one place.
That netted out to:
```
  publishing {
     publications {
        mavenJava(MavenPublication) {
           from components.java
           // specify dependencies like: org.apache.edgent:edgent.api.topology:0.4.0
           groupId = build_group
           artifactId = "${project.group}.${project.name}" 
           artifact sourceJar
        }
     }
  }  
```
Does that naming scheme / impl seem good to you?

xuanyue202: sounds good.

xuanyue202: and I think the next step that we all look farward to is to make Edgent go into Maven Central Repository :)

dlaboss:
Going to merge this (thanks again) and will tweak afterwards.

PR #182, opened 2016-08-29
tidy up publishToMavenLocal support

dlaboss:
- publish groupId is org.apache.edgent
- publish war when applicable


dlaboss:
An unrelated test failure.  Looks like a timing sensitivity in IotProviderTest.  Creating a Jira for it.
Trying close/reopen to cycle again.

PR #183, opened 2016-08-29
EDGENT_243 console legend wider

home4slc:
Added some extra space around the legend box in the console due to the longer package names after changing from apache.quarks to org.apache.edgent.

### DIFF
--- a/console/servlets/webapp_content/html/index.html
+++ b/console/servlets/webapp_content/html/index.html
@@ -58,12 +58,12 @@
 <button id="toggleTimer" title='Pause graph' type="button">Pause graph</button>
 </div>
 
-<div style='width: 300px; margin-left: 30px; clear:both;'>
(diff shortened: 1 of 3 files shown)


dlaboss:
wondering... did you consider omitting the "org.apache.edgent" or such?  Probably 99% of all oplets will be edgent oplets so the prefix is really just undesired noise imo.
Related, an oplet with a long package name (e.g., TextFileReader; can see with scripts/connectors/file sample) runs off the right edge of the box; and there isn't a right edge drawn on my mac -- that's w/o this PR's changes.

home4slc:
That is certainly an option.  My thoughts were that if someone had a package that was not org.apache.edgent that I the legend box would need to accommodate that length anyway.

If folks think that it is more desirable to remove org.apache.edgent then it can be done.

Cheers,

Susan

PR #190, opened 2016-09-14
Edgent-240 [gradle] Overhaul dependency specification and handling

dlaboss:
The upshot is to do like the ant-based scripts and specify
inter-project, core-ext, and project-private-ext
dependencies against jars in the target dir.
The original conversion specified on inter-project dependencies on
"projects" (hence their artifacts not in the target-dir) and that caused
a host of issues.

With the changes things like manifest-classpath generation are
simplified and overall the gradle build now adhears to the policies:
- Edgent projects depend on other project's jars in the target-dir, not
their classes
- Edgent project jars have a manifest-classpath that handles the
project's "private" inter-project dependencies as well as the project's
"private" external component dependencies.
- We build a target dir that includes the project's jar as well as the
project's external dependency jars
- The tests compile and run against the project jars in the target dir
(as external/integration test code would).
- The samples compile and run against the project jars in the target dir
(as user code would).

All seems fine including when building and testing with
ANDROID_SDK_PLATFORM and DERBY_HOME set:
- all tests pass (sans manual run of kafka tests)
- sample scripts seem to run appropriately
- console comes up

TODOs
- edgent.android.{topology,hardware}.jar manifest classpath and the
"android.jar" - compared to ant-build?

More details of the changes...

- define addTargetDirProjectJarDependency(),
addTargetDirExtJarDependency(), addTargetDirCoreExtDependency(),
addTargetDirCoreExtDependencies(), addProjectExtDependency()
- change all build.gradle to use them
- configure common sample dependencies in root build.gradle
- adapt manifest classpath generation in light of above
- rework how common_ext_dependencies files get added to target-dir/ext

Other cleanup,etc
- generate project jar/war directly into targetdir with the correct name
(remove "copy/rename" of jar created by default far task)
- fix ant -> gradle conversion of api/function project
- add previously (benign) absent api/function dependency from
api/topology
- remove updateTargetDir task -- just add processing as "doLast" to the
assemble task
- other root build.gradle refactoring for hygiene
- fix issues caused by kafka's transitive dependencies
- update connectors:iotp to watson-iot:0.1.5 - initial motivation was to
cleanup the fact that 0.1.1 had dependency on a mqttv3 SNAPSHOT version
- fix aggregateJavadoc processing warnings
- fix test runtime classpath so it uses the project's target-dir jar

### DIFF
--- a/analytics/math3/build.gradle
+++ b/analytics/math3/build.gradle
@@ -12,10 +12,12 @@
  * limitations under the License.
  */
 dependencies {
-  compile project(':api:topology')
-  compile 'org.apache.commons:commons-math3:3.4.1'
-  compile core_ext_dependencies
-  testCompile project(':providers:direct')
+  addTargetDirProjectJarDependency 'compile', ':api:topology'
+  addProjectExtDependency 'compile', 'org.apache.commons:commons-math3:3.4.1'
+  addTargetDirCoreExtJarDependencies 'compile'
+  
+  addTargetDirProjectJarDependency 'testCompile', ':providers:direct'
+  // N.B. root project adds test common dependencies
 }
 
 addCompileTestDependencies ':api:topology', ':providers:direct'
--- a/analytics/sensors/build.gradle
+++ b/analytics/sensors/build.gradle
@@ -12,9 +12,11 @@
  * limitations under the License.
  */
 dependencies {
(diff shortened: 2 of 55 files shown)


dlaboss:
unrelated ci failure, rerun

PR #194, opened 2016-09-22
Edgent-240 [gradle] use gradle from travis-ci runs

dlaboss:


dlaboss:
unrelated mqtt/websocket test failures. try for a clean run.

PR #199, opened 2016-09-27
Edgent-254 [gradle] doc gradle as the default env

dlaboss:


dlaboss:
unrelated ci test failure for test7

PR #200, opened 2016-09-28
Edgent-255 improve StreamScope.setEnabled() behavior

dlaboss:


dlaboss:
created EDGENT-256 for unrelated test failure

PR #214, opened 2016-10-07
EDGENT-262 console license changes

home4slc:
I added a console/servlets/LICENSE file containing the following:

This initial line of text:
"This product contains content covered by MIT/BSD-3-Clause licenses."
then where the ext files came from, and their associated license type, a URL to the license or path in the repository to the license, and added a couple of license.
I also updated <repo>/LICENSE to mention the console/servlets/LICENSE file

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,6 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+   ======================================
+   This product contains content covered by MIT and "BSD 3 Clause" licenses. See console/servlets/LICENSE.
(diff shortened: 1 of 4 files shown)


dlaboss:
+1 The new `servlets/LICENSE` and change to `/LICENSE`.

-1 But if we retain the *.LICENSE files (e.g., `sankey.d3.BSD.LICENSE`) I think their text needs to be the same as the text at the URL noted in `servlets/LICENSE` file.   But I have an alternative suggestion that might simplify things:

Since `servlets/LICENSE` has the URLs to the corresponding licenses, add a http URL for the MIT license (`https://opensource.org/licenses/MIT`) and remove all of the *.LICENSE files under js/ext.

How does that sound?  (sorry for previously asking to add the MIT license file to the repo)

home4slc:
That is fine.  I'll do this:

1) remove all of the *License* files under /ext 
2) Only point to the licenses via the http URL

Susan

PR #229, opened 2016-11-07
Fix publish by ignoring platform projects

xuanyue202:
This commit fixed the following issue:

In the current version, if you run gradle publishToMavenLocal or gradle publish, it will throw the following error:

Execution failed for task ':platform:android:publishMavenJavaPublicationToMavenLocal'.
> Failed to publish publication 'mavenJava' to repository 'mavenLocal'
   > Invalid publication 'mavenJava': artifact file does not exist: '/git/quarks/incubator-edgent/build/distributions/java8/platform/android/lib/edgent.platform.android.jar'

The edgent.platform.java project also has the similar problem that prevent the project from publishing to Maven.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -210,8 +210,10 @@ subprojects {
   if (aggregatorOnlyProjects.contains(project.path)) {
     return
   }
-
-  apply plugin: 'maven-publish'
+  
+  if(!project.group.equals("edgent.platform")){
+  	apply plugin: 'maven-publish'
+  }
   apply plugin: 'java'
   apply plugin: "jacoco"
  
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -621,6 +623,7 @@ subprojects {
 
   // support for 'gradle publishToMavenLocal' etc 
   // TODO publishing test.{fvt,svt} and samples ... doesn't seem desirable? e.g., we're excluding test.{fvt,svt} jars from the tgz
+  if(!project.group.equals("edgent.platform")){
```
dlaboss: Thanks for your contribution!
Rather than duplicate the "edgent.platform" specific conditionalization here, instead, how about making it more general?
```if (project.pluginManager.hasPlugin('publishing')) {```

At a higher level, I think I wish there were a way to address the problem without project-specific conditionalization in the root build.gradle.  Do you have any thoughts on an approach where specific sub projects build.gradle (like platform/*/build.gradle) could do something to override the "typical/default subproject" behavior in the root build.gradle subprojects processing?  e.g., doing something like setting a "omitPublishingPlugin" property (that the "default" subproject processing checks) or somehow nullifying the "default" publishing configuration that's created. ???

dlaboss: ping @xuanyue202

PR #242, opened 2016-11-15
Add pr-237 to Release1.0.0 branch

Cazen:


Cazen:
Hmm I'd just merge previous pr-237 to 1.0.0 branch but looks another commit(232f89c) included in this PR.
I checked the "Files changed" and it looks ok. But if there are some problem, I will recreate PR with write them in a direct way instead of using merge

PR #256, opened 2016-12-21
[Edgent-325] improve jdbc, kafka, and java7 test doc

dlaboss:


dlaboss:
unrelated timing sensitive test failure w/java7  <testcase name="keyedTimeBatchWindowTest" classname="org.apache.edgent.test.window.WindowTest" time="11.011">

PR #280, opened 2017-02-01
[Edgent-377] IotGateway

dlaboss:
Add IotGateway
Add IotDevice.{getDeviceType(),getDeviceId(),CMD_DEVICE}

These IotDevice changes break existing IotDevice implementations but in
practice this won't be a problem - its highly unlikely any IotDevice
implementations exist outside of this repository (and adapting is
trivial).  "default" can't be used due to support for Java7/Android.  

Other schemes are possible that avoid this but they don't model things
as well (an IotDevice really does have an Id), introduce even more types
(e.g., a IotGatewayConnectedDevice), and/or are less convenient to use.
e.g., those new IotDevice methods could be migrated to IotGateway -
IotGateway.getDeviceId(IotDevice).

### DIFF
--- a/connectors/iot/src/main/java/org/apache/edgent/connectors/iot/IotDevice.java
+++ b/connectors/iot/src/main/java/org/apache/edgent/connectors/iot/IotDevice.java
@@ -29,6 +29,22 @@ import com.google.gson.JsonObject;
 
 /**
  * Generic Internet of Things device connector.
+ * <p>
+ * IotDevice characteristics:
+ * <ul>
+ * <li>{@code IotDevice.getDeviceTypeId()} returns an opaque value whose form
+ *      is the domain of an IoT connector implementation.</li>
+ * <li>{@code IotDevice.getDeviceId()} returns an opaque value whose form
+ *      is the domain of an IoT connector implementation.
+ *      The value is unique for a particular a logical device.
+ * <li>{@code IotDevice.equals()} returns true if two IotDevice instances are
+ *     for the same logical device, false otherwise.</li>
+ * <li>{@code IotDevice.hashCode()} returns the same value for all IotDevice instances
+ *     for the same logical device.</li>
+ * <li>{@code IotDevice} may be used as a {@link org.apache.edgent.topology.TWindow TWindow} partition key.</li> 
+ * </ul>
+ * 
+ * @see IotGateway
  */
 public interface IotDevice extends TopologyElement {
     
@@ -36,6 +52,20 @@ public interface IotDevice extends TopologyElement {
      * Device event and command identifiers starting with {@value} are reserved for use by Edgent.
      */
     String RESERVED_ID_PREFIX = "edgent";
+    
+    /**
+     * Get the device's opaque device type identifier.
+     * TODO remove the "default" - avoids compilation errors while discussing this.
+     * @return
+     */
+    public default String getDeviceType() { return "a-device-type-id"; }
+    
+    /**
+     * Get the device's unique opaque device identifier.
+     * TODO remove the "default" - avoids compilation errors while discussing this.
+     * @return
+     */
+    public default String getDeviceId() { return "a-device-id"; }
 
     /**
      * Publish a stream's tuples as device events.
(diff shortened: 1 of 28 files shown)


ddebrunner:
Which issue is this, doesn't seem to be 374.

connectors/iot/src/main/java/org/apache/edgent/connectors/iot/IotDevice.java
```diff
@@ -29,13 +29,43 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+ * <li>{@code IotDevice.getDeviceTypeId()} returns an opaque value whose form
+ *      is the domain of an IoT connector implementation.</li>
+ * <li>{@code IotDevice.getDeviceId()} returns an opaque value whose form
+ *      is the domain of an IoT connector implementation.
+ *      The value is unique for a particular a logical device.
+ * <li>{@code IotDevice.equals()} returns true if two IotDevice instances are
+ *     for the same logical device, false otherwise.</li>
+ * <li>{@code IotDevice.hashCode()} returns the same value for all IotDevice instances
+ *     for the same logical device.</li>
+ * <li>{@code IotDevice} may be used as a {@link org.apache.edgent.topology.TWindow TWindow} partition key.</li>
```
ddebrunner: Curious why this is here. If IoTDevice holds a connection to send/receive messages and it doesn't seem like that should be partition key.

dlaboss:
woops fixed the title - EDGENT-377

ddebrunner:
Found it, https://issues.apache.org/jira/browse/EDGENT-377

connectors/iot/src/main/java/org/apache/edgent/connectors/iot/IotGateway.java
```diff
@@ -0,0 +1,122 @@
...
+   * This is logically equivalent to a union of a collection of individual IotDevice specific
+   * command streams but enables an IotGateway implementation to implement it more efficiently. 
+   * 
+   * @param devices
+   *            Only return commands for the specified connected devices
+   * @param commands Command identifiers to include. If no command identifiers are provided then the
+   * stream will contain all device commands for the specified devices.
+   * @return Stream containing device commands.
+   */
+  TStream<JsonObject> commands(Collection<IotDevice> devices, String... commands);
```
ddebrunner: I think this should be `Set<String> deviceIds` not a collection of `IotDevice`.

ddebrunner:
I think I should be able to use/implement an IoTGateway without having to have `IoTDevice` instances for each connected device.

As an example the gateway may already be routing device events/commands as part of its core functionality so I don't need connectivity in Edgent to the devices, instead I just want to write an analytics app that analyzes the events/ commands going through the gateway.

connectors/iot/src/main/java/org/apache/edgent/connectors/iot/IotDevice.java
dlaboss: Relates to the use of an IotDevice instance as an API object representing an IoT device.  More on that in the response to your other usage model question.

dlaboss:
The use of IotDevice instances in an IotGateway context was to preserve / extend the existing API model - “an IotDevice represents a particular IoT device and you use an instance of it to publish events / subscribe to commands for a specific IoT device”.

You’d use IotGateway.getIotDevice(…) to get an IotDevice for a particular connected device and then just use it as you would in a non-gateway context.

Sounds like you prefer an alternate model for the gateway context that only uses a simple device-identifier string and events/commands methods that take explicit device-ids as arguments.

```
IotGateway extends IotDevice {
  commandsForDevice(device-id, …)  // can’t be 	commands(device-id-string, …)
  commandsForDevice(Set<device-id>, …)  // empty set -> no filtering / include all? gateway too?
  commandsForType(type-id, …)
  eventsForDevice(device-id, TStream,…)
  eventsForDevice(device-id-fn, TStream, event-id-fn,…)

  String getDeviceId(Map<String,String> impl-device-id-attrs)
  String getDeviceType(String deviceId)
}
```
Does that capture it?  I can live with that.

Probably doesn’t change your mind about but I noticed an omission in the original proposal for publishing device specific events from a multi-device stream:
      `events(Function<JsonObject,IotDevice> device-id-fn, TStream<JsonObject>, …)`

ddebrunner:
No, that's not what I meant. 

I meant that I should be able to use a gateway, without having to create IoTDevices to perform analytics on connected device events/commands. Say I don't want to interact with the devices, just analyze what is going on with them.

dlaboss:
Sorry, I'm not following you.  It may have been clear to all but just to be sure... I've only been talking about an Edgent app on a gateway device _as a client of the IoT hub_ - how the gateway app publishes events to the hub as if they were from a "connected device" (a device serviced by the gateway) and how the gateway app receives commands from the hub that are for a connected device.  

IotGateway.getIotDevice(...) wouldn't/shouldn't engage in any conversation with the hub.  It just gives the gateway app a handle representing the connected device's identity (which can be accessed via IotDevice.getDeviceId()) and the ability to publish events and/or to subscribe to events for that device.

(Additionally, a gateway / the gateway app can publish device events to the hub from "the gateway device id" and subscribe to device commands for "the gateway device id".)

This topic/feature doesn't say anything about interactions from an Edgent gateway app with the gateway's connected devices.

Where does that leave us?  Is the original proposal satisfactory for the above?  Please elaborate on the scenario you're talking about, how it's not satisfied by the above, what specific changes to the proposal are needed to address it.  Thanks!

dlaboss:
Lacking any further feedback / comments I went ahead with an implementation.

dlaboss:
looks like travis may be wedged. closing / reopening to try to clear things up.

PR #303, opened 2017-03-30
[Edgent-406] improve ApplicationServiceMXBean.registerJar() doc

dlaboss:


dlaboss:
The mqtt test failures are independent of the changes (javadoc only in unrelated classes).

PR #309, opened 2017-06-02
Migration of the Edgent project to be able to build with Apache Maven

chrisdutz:
I finally managed to solve all problems and untangle all knots I wasn't that happy with.

The current version should be buildable with a simple "mvn package"
This would include running all unit-tests as part of the build.


analytics/math3/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.edgent.analytics</groupId>
+    <artifactId>analytics</artifactId>
+    <version>1.2.0-SNAPSHOT</version>
```
dlaboss: I assume this is just a temporary situation and a variable will be defined once for the Edgent version being built and then just referenced everywhere else?

analytics/math3/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+      <artifactId>api-topology</artifactId>
+      <version>1.2.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-math3</artifactId>
+      <version>3.4.1</version>
+    </dependency>
+
+    <dependency>
```
dlaboss: newbie question: I see this in many (all?) pom.xml:  Why are two dependency declarations needed for `providers-direct`?  Why isn't just the 2nd one sufficient (like the one for `api-topology`)?

analytics/math3/pom.xml
chrisdutz: Actually not. Especially the parent is the only place where a variable can't be used. But in maven the version number of the all artifacts in the same reactor (collection of maven modules in one build) is usually explicitly written as the maven-release-plugin takes care of updating all the version numbers. The only place it tends to have problems with is when referencing versions in the plugin configuration. That's why I used variables in the maven-dependency-plugin configuration.

analytics/math3/pom.xml
chrisdutz: Well it actually is handling two different artifacts ... one is the main jar (first declaration) which contains the productive code which is added to the compile classpath. The second adds the jar containing the test-code (type=test-jar) to the test classpath (scope=test)

dlaboss:
Please change "/target" to "target/" in the top level `.gitignore`.  `git status` is showing the per-module generated artifacts. Thanks!

dlaboss:
With the merge from upstream, in the top level pom.xml's java7 Profile (and retrolambda config), can the following now be removed?

```
line 565
    <configuration>
      <defaultMethods>true</defaultMethods>
    </configuration>
```

platforms/android/connectors/pom.xml
```diff
@@ -37,19 +37,19 @@
...
     <module>iotp</module>
-    <module>javax.websocket-client</module>
-    <module>javax.websocket-server</module>
     <module>jdbc</module>
     <module>kafka</module>
     <module>mqtt</module>
     <module>pubsub</module>
     <module>serial</module>
-    <module>wsclient</module>
+    <module>websocket-base</module>
```
dlaboss: missing <module>websocket</module>

connectors/websocket/pom.xml
```diff
@@ -27,10 +27,10 @@
     <version>1.2.0-SNAPSHOT</version>
   </parent>
 
-  <artifactId>edgent-connectors-websocket-wsclient</artifactId>
```
dlaboss: unrelated to this change I see that j8 and j7 websocket/pom.xml declare a dependency on the direct provider.  That can't be correct. A connector can't depend on a provider.

chrisdutz: Ok ... it seems that there is a required dependency to "direct" provider, but I moved the dependency from a "compile" to a "test" dependency as removing "direct" completely caused compilation errors in tests.

platforms/android/connectors/pom.xml
chrisdutz: Added the module in 6cb9c73

pom.xml
```diff
@@ -223,6 +223,15 @@
             </lifecycleMappingMetadata>
           </configuration>
         </plugin>
+
+        <!-- Inject the "incubating" into the output filename -->
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-assembly-plugin</artifactId>
+          <configuration>
+            <finalName>edgent-incubating-${project.version}</finalName>
```
dlaboss: Just checking that omitting "-parent" was intentional. (I'm not sure what the best-practice is for this)

chrisdutz: It was intentional ... The POM is named edgent-parent, but the source release contains the entire project. That's why I omitted the "parent" from the name.

chrisdutz:
Type= jar is the default. So if you omit the type, that's what have will use. So a type declaration would be redundant.

pom.xml
dlaboss: Makes sense, thanks.  I also noticed that the current bundles have an apache prefix ("apache-edgent-<ver>....tgz").  Don't recall if that's a best practice, I seem to have a recollection of Justin noting something about it :-)  Your thoughts?   

We also currently create .md5 and .sha files for distribution.  Are they missing or do they only get created as part of a different task (I used mvn install -Papache-release)?  

Previously we only created a tgz bundle.  Now it looks like only a zip?  Seems OK but wanted to double check.

chrisdutz: Yeahh ... remember having that discussion with him for Flex ... Just added an "apache-" and He'll be happy.

console/servlets/pom.xml
```diff
@@ -49,23 +56,60 @@
...
-                <artifactId>jquery-ui</artifactId>
-                <version>1.11.4</version>
-            </dependency>
-            <dependency>
-                <groupId>org.webjars</groupId>
-                <artifactId>jquery</artifactId>
-                <version>1.11.2</version>
-            </dependency>
-            -->
+            <!-- TODO: Still got one unlicensed file in there ... -->
```
dlaboss: Please update the top level source LICENSE file to reflect only what's now bundled in the source.
Presumably the WAR's included LICENSE continues to require noting that it bundles d3,jquery,...?

connectors/kafka/pom.xml
```diff
@@ -42,7 +42,12 @@
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka_2.10</artifactId>
       <version>0.8.2.2</version>
-      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
```
dlaboss: I just verified that each of these non-edgent deps need the non-transitive specification (not just kafka_2.10).  Without it, copy-dependencies (and uber jar) pull in unwanted things relative to the gradle/ant build:

Only in DEPS-WITH-KAFKA-TOO-MUCH/: jline-0.9.94.jar
Only in DEPS-WITH-KAFKA-TOO-MUCH/: lz4-1.2.0.jar
Only in DEPS-WITH-KAFKA-TOO-MUCH/: netty-3.7.0.Final.jar
Only in DEPS-WITH-KAFKA-TOO-MUCH/: slf4j-log4j12-1.6.1.jar
Only in DEPS-WITH-KAFKA-TOO-MUCH/: snappy-java-1.1.1.7.jar

dlaboss:
Please add the following dependency to connectors/jdbc.  With it, the JDBC connector tests then pass and they will also run with the rest of the tests.  Manual setup/run not needed :-)

```
<!-- https://mvnrepository.com/artifact/org.apache.derby/derby -->
<dependency>
    <groupId>org.apache.derby</groupId>
    <artifactId>derby</artifactId>
    <version>10.13.1.1</version>
    <scope>test</scope>
</dependency>
```

connectors/kafka/pom.xml
dlaboss: Added a commit to address this.

dlaboss:
note: above above commits addressed my issues/comments above (or course you'll have to pull the feature branch to update your clone's content to include them)

dlaboss:
Chris, I just committed the first round changes for the samples (above).
The desire was to create a separate source release bundle for the samples (perhaps a precursor to migrating them to their own repository at some point). The samples pom's do not have the edgent-parent pom as a parent.  I wasn't able to figure out how to exclude the sample sources from the "core" apache-edgent source bundle.  The sample bundle is built via `cd samples; mvnw clean install -papache-release`
See comments the edgent-parent/pom.xml related to "samples".
See samples/README.md.
At this moment only topology/run-sample.sh is present.  All non-console-non-http-server topology samples should run successfully.  TODO - other sample categories.

dlaboss:
The Edgent console page now comes up in the browser, but there's still a problem.  Following its "loading graph" progress message, no topology graph is rendered.  Reproduce via the topology DevelopmentSample sample.  Is the generated uber jar or `servlets.war` is missing something?

```
cd samples
./mvnw clean install
cd topology
./run-samples.sh DevelopmentSample
# open browser to the "http://localhost:<port>/console" that's emitted
```

dlaboss:
Chris, noticed building (now?) reports a warning about duplicate declaration of the maven-assembly-plugin in pom.xml.  There are in fact two decls of it.  Also just a reminder of Edgent Console problem noted above.  Thanks

dlaboss:
Chris, when importing into Eclipse there's one residual warning for the top pom.xml: `Overriding managed version 3.6 for maven-site-plugin`.  The pom is specifying ver 3.4.  If that's intentional can a clarification as to why not 3.6 be added to the file and as well as an annotation to squelch the warning: `<version>3.4</version><!--$NO-MVN-MAN-VER$-->`.  Thanks.

analytics/math3/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,12 @@
...
+Copyright 2017 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+===============================================================================
+
+Portions of this bundle were developed by IBM Corp.
+Copyright IBM Corp. 2015, 2016
+
```
dlaboss: Yikes... that's pretty cumbersome :-)  
"Copyright 2017 The Apache..." needs to be "Copyright 2016-2017...", right?
And we'll need to remember to change them all the first release of a year.

dlaboss: Another issue? ...the original binary bundle jar NOTICE (in licenses/ibm-contrib-notice) includes an incubating disclaimer sentence. It's omitted in these.  I suspect it's required.

chrisdutz: Well the option I would probably go for is to make a resource bundle like the one used in the remote resources bundle. The year could be automatically inserted by maven ... in the worst case it would be just a property on the parent project. This is just the quick fix for the open issue. I used the content of the official template the apache foundation provides for incubating projects. The incubating message is in the Disclaimer and not in the Notice for these. If this were required, probably the Apache Foundation would have updated those templates. And as Justin said "Keep it as short as possible" :-)

dlaboss: I didn't realize it wasn't present in the official template... excellent!

dlaboss: oh... but we still a starting date of 2016 before we can release this

chrisdutz:
All Changes have been merged back to the original edgent repo, but to the develop branch instead of the master I created the pull-request for.

dlaboss:
WooHoo!  I noticed it's 5 months to the day (and 289 commits) from the first commit on this effort. whew/yikes :-)

PR #348, opened 2018-03-04
[EDGENT-447][Connectors] contribute a RabbitMQ connector

yanghua:
## What is the purpose of the change

This PR contribute a RabbitMQ connector for edgent.

## Brief change log

  - *Implement a `RabbitmqProducer` and `RabbitmqConsumer` as a source & sink to integrate with edgent*
  - *Implement a test case to verify the connector*

## Verifying this change

This change added tests and can be verified as follows:

  - *install & start RabbitMQ server then run `RabbitmqStreamsTestManual.testSimple` it will pass throgh*

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes)
  - The public API, i.e., is any changed : (no)
  - Anything that affects deployment: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (documented with JavaDoc)

### DIFF
--- a/connectors/pom.xml
+++ b/connectors/pom.xml
@@ -49,6 +49,7 @@
     <module>websocket-jetty</module>
     <module>websocket-misc</module>
     <module>websocket-server</module>
+    <module>rabbitmq</module>
   </modules>
 
 </project>
--- a/connectors/rabbitmq/pom.xml
+++ b/connectors/rabbitmq/pom.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>edgent-connectors</artifactId>
+    <groupId>org.apache.edgent</groupId>
+    <version>1.3.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>edgent-connectors-rabbitmq</artifactId>
+
+  <name>Apache Edgent (Java 8): Connectors: RabbitMQ</name>
+
+  <properties>
+    <remote-resources-maven-plugin.remote-resources.dir>../../src/main/ibm-remote-resources</remote-resources-maven-plugin.remote-resources.dir>
+    <edgent.version></edgent.version>
+    <rabbitmq.version>5.1.2</rabbitmq.version>
+  </properties>
+
+  <dependencies>
+    <!-- edgent start -->
+    <dependency>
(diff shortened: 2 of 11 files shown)


yanghua:
Hi @chrisdutz could you please review this PR? Thanks!

chrisdutz:
Hi, so I just did a quick check of the new dependency. It's tripple licensed and one of the available licenses is Apache 2.0, so we're good on the side I was slightly concerned about. Will check out the changes in a little more detail now.

dlaboss:
I know this has already been merged.  Nice Job.  Couple of high level questions:

- Why not follow the MQTT connector's MqttStreams API style (have RabbitmqStreams) instead of making the main API be Rabbitmq{Consumer,Producer}?  (note there's been a recent discussion on @dev about how that Kafka connector's Consumer/Producer naming style causes confusion)

- Rather than introduce the RabbitmqConfigKeyConstants I wonder if it would be better to require the connector user to just construct and pass in a com.rabbitmq.client.ConnectionFactory?  The connector would have a smaller footprint / less "wrapper" to doc and maintain.  I know the MQTT connector defined a MqttConfig class but in retrospect I wonder if that was a mistake. I think the idea at the time was to make the MQTT connector's API independent of using the paho MQTT code under the covers.  That level of isolation seems unnecessary for the RabbitMQ connector.

- Does this version of the connector survive transiently broken connections and/or delivery failures, or initial connection failures?

- Can creating/starting/stopping a RabbitMQ server instance be embedded in the test code?  Or is there a public RabbitMQ server instance available? That way the tests could be included as part of the automated tests.  If not / until it is, then it seems like DEVELOPMENT.md should be augmented with information for the RabbitMq connector like the Kafka connector.

- Your thoughts about adding a RabbitMQ connector sample to the samples https://github.com/apache/incubator-edgent-samples

yanghua:
Hi Dale,

Thanks for your suggestion.  The answers :


   - Why create two classes (RabbitmqConsumer and RabbitmqProducer), here I
   followed Kafka connector and they should not share the same rabbitmq
   connection and channel. I also answered in that disscuss mail about naming
   style
   - About the RabbitmqConfigKeyConstants : the reason is the rabbitmq
   client need many config arguments, I also followed the Kafka connector used
   a Map<String, Object> as the argument box. I think we should not expose the
   RabbitMQ client's ConnectionFactory expose to user, it's not safety  and
   user will need more knowledge about RabbitMQ's client. For there many
   arguments, the *builder pattern* would be better.
   - The RabbitMQ client can handle the connection failure and specific
   problem with *enable recovery* and *recovery interval* setting.
   - The same as Kafka it do not support automated tests. I have done the
   test. If anyone want to run the test case, the rabbitmq server should be
   started manually. All the description (about installation and launch steps)
   I have written in the test class. And there is no public rabbitmq server
   could be used as test server.
   - Yes,  the usage of this connector should be added to samples
   collection, I will do this in the future.

Vino yang
Thanks.


2018-03-23 3:13 GMT+08:00 Dale LaBossiere <notifications@github.com>:

